PR #1, opened 2018-03-27
Feature/update seeking and session initialization

offnertho:
Initially it was not really possible to update the session data during data collection. Therefore we had to wait with initializing the session, until all the required data was available.

Now we initialize the session right away, and then update it, when the data is available. 

Additionally adjusted the setting of the player states.

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -64,6 +64,9 @@ export class ConvivaAnalytics {
         + 'Please load the Conviva script (conviva-core-sdk.min.js) before Bitmovin\'s ConvivaAnalytics integration.');
       return; // Cancel initialization
     }
+
+    // player versions <=7.2 did not have a ON_PLAYING event
+    // we track this change to correctly transition to the playing state
     this.hasPlayingEvent = Boolean(player.EVENT.ON_PLAYING);
 
     // Assert that this class is instantiated before player.setup() is called.
@@ -165,18 +168,18 @@ export class ConvivaAnalytics {
   }
 
   private updateSession = () => {
+    this.updateContentMetadata();
+    this.playerStateManager.updateContentMetadata(this.contentMetadata);
+  };
+
+  private updateContentMetadata() {
     let source = this.player.getConfig().source;
-    // Create a ContentMetadata object and supply relevant metadata for the requested content.
 
     if (this.contentMetadata.assetName !== this.getAssetName(source)) {
       this.contentMetadata.assetName = this.getAssetName(source);
     }
 
     this.contentMetadata.viewerId = source.viewerId || this.config.viewerId || null;
-
-    this.contentMetadata.duration = 0; // TODO how to handle HLS Chrome deferred duration detection?
-    this.contentMetadata.streamType = Conviva.ContentMetadata.StreamType.UNKNOWN; // TODO how to handle HLS deferred
-                                                                                  // live detection?
     this.contentMetadata.streamUrl = this.getUrlFromSource(source);
     this.contentMetadata.custom = {
       'playerType': this.player.getPlayerType(),
@@ -184,10 +187,9 @@ export class ConvivaAnalytics {
       'vrContentType': this.player.getVRStatus().contentType,
     };
     this.contentMetadata.duration = this.player.getDuration();
-    this.contentMetadata.streamType = this.player.isLive() ? Conviva.ContentMetadata.StreamType.LIVE // TODO how to handle HLS deferred live detection?
-      : Conviva.ContentMetadata.StreamType.VOD;
-    this.playerStateManager.updateContentMetadata(this.contentMetadata);
-  };
+    this.contentMetadata.streamType
+      = this.player.isLive() ? Conviva.ContentMetadata.StreamType.LIVE : Conviva.ContentMetadata.StreamType.VOD;
+  }
 
   private getAssetName(source: any): string {
     let assetName;
@@ -225,12 +227,10 @@ export class ConvivaAnalytics {
       return;
     }
 
-    // Do not start a new session when a session is already existing
-    // Happens after ad playback, when the actual source is restored and an ON_SOURCE_LOADED event issued. Because
-    // we suppress the ON_SOURCE_UNLOADED event which unloads the temporary ad source, we must also ignore this
-    // event. By ignoring these ad-induced events, we end up with a clean ON_SOURCE_LOADED/ON_SOURCE_UNLOADED
-    // sequence which only concerns the actual source.
-    this.updateSession();
+    // in case a source has been loaded after an source_unloaded initialize a new session
+    if (!this.isValidSession()) {
+      this.initializeSession();
+    }
   };
 
   private onReady = (event: any) => {
@@ -256,9 +256,14 @@ export class ConvivaAnalytics {
 
   private onPlay = (event: any) => {
     this.debugLog('play', event);
-    this.updateSession();
+
+    // in case the playback has finished and the user replays the stream create a new session
+    if (!this.isValidSession()) {
+      this.initializeSession();
+    }
+
     if (!this.hasPlayingEvent) {
-      this.onPlaybackStateChanged();
+      this.updateSession();
     }
   };
 
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -1,10 +1,10 @@
 ///<reference path="Conviva.d.ts"/>
-import {Html5Time} from './Html5Time';
-import {Html5Timer} from './Html5Timer';
-import {Html5Http} from './Html5Http';
-import {Html5Storage} from './Html5Storage';
-import {Html5Metadata} from './Html5Metadata';
-import {Html5Logging} from './Html5Logging';
+import { Html5Time } from './Html5Time';
+import { Html5Timer } from './Html5Timer';
+import { Html5Http } from './Html5Http';
+import { Html5Storage } from './Html5Storage';
+import { Html5Metadata } from './Html5Metadata';
+import { Html5Logging } from './Html5Logging';
 
 export declare type Player = any; // TODO use player API type definitions once available
 
@@ -268,7 +268,7 @@ export class ConvivaAnalytics {
   };
 
   private onSeek = (event: any) => {
-    this.playerStateManager.setPlayerSeekStart(event.seekTarget * 1000);
+    this.playerStateManager.setPlayerSeekStart(Math.round(event.seekTarget * 1000));
   };
 
   private onSeeked = () => {
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -42,6 +42,7 @@ export class ConvivaAnalytics {
   private config: ConvivaAnalyticsConfiguration;
   private contentMetadata: ContentMetadata;
   private hasPlayingEvent: boolean;
+  private sessionDataPopulated: boolean;
 
   private systemFactory: Conviva.SystemFactory;
   private client: Conviva.Client;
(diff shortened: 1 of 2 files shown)


src/ts/ConvivaAnalytics.ts
```diff
@@ -130,51 +135,78 @@ export class ConvivaAnalytics {
...
     }
 
+    this.playerStateManager.setPlayerState(Conviva.PlayerStateManager.PlayerState.STOPPED);
     this.client.attachPlayer(this.sessionKey, this.playerStateManager);
     this.debugLog('startsession', this.sessionKey, event);
+  }
+
+  private updateSession = () => {
+    let source = this.player.getConfig().source;
+    // Create a ContentMetadata object and supply relevant metadata for the requested content.
```
protyposis: Do you think this comment makes sense here? What does it comment, the following empty line?

src/ts/ConvivaAnalytics.ts
```diff
@@ -130,51 +135,78 @@ export class ConvivaAnalytics {
...
+
+    if (this.contentMetadata.assetName !== this.getAssetName(source)) {
+      this.contentMetadata.assetName = this.getAssetName(source);
+    }
+
+    this.contentMetadata.viewerId = source.viewerId || this.config.viewerId || null;
+
+    this.contentMetadata.duration = 0; // TODO how to handle HLS Chrome deferred duration detection?
+    this.contentMetadata.streamType = Conviva.ContentMetadata.StreamType.UNKNOWN; // TODO how to handle HLS deferred
```
protyposis: You were working on improvements regarding reporting of `duration` and `streamType`. Do these `TODO`s here still make sense? Is the reporting still not working correctly? Or does it maybe only not work under certain conditions that could be documented somewhere?

src/ts/ConvivaAnalytics.ts
```diff
@@ -130,51 +135,78 @@ export class ConvivaAnalytics {
...
+                                                                                  // live detection?
+    this.contentMetadata.streamUrl = this.getUrlFromSource(source);
+    this.contentMetadata.custom = {
+      'playerType': this.player.getPlayerType(),
+      'streamType': this.player.getStreamType(),
+      'vrContentType': this.player.getVRStatus().contentType,
+    };
+    this.contentMetadata.duration = this.player.getDuration();
+    this.contentMetadata.streamType = this.player.isLive() ? Conviva.ContentMetadata.StreamType.LIVE // TODO how to handle HLS deferred live detection?
```
protyposis: Why do you set `duration` and `streamType` twice in the same method?

src/ts/ConvivaAnalytics.ts
```diff
@@ -186,89 +218,73 @@ export class ConvivaAnalytics {
...
 
-    if (this.isValidSession()) {
-      // Do not start a new session when a session is already existing
-      // Happens after ad playback, when the actual source is restored and an ON_SOURCE_LOADED event issued. Because
-      // we suppress the ON_SOURCE_UNLOADED event which unloads the temporary ad source, we must also ignore this
-      // event. By ignoring these ad-induced events, we end up with a clean ON_SOURCE_LOADED/ON_SOURCE_UNLOADED
-      // sequence which only concerns the actual source.
-      return;
-    }
+    // Do not start a new session when a session is already existing
```
protyposis: Does this comment still make sense?

src/ts/ConvivaAnalytics.ts
```diff
@@ -186,89 +218,73 @@ export class ConvivaAnalytics {
...
+    // Happens after ad playback, when the actual source is restored and an ON_SOURCE_LOADED event issued. Because
+    // we suppress the ON_SOURCE_UNLOADED event which unloads the temporary ad source, we must also ignore this
+    // event. By ignoring these ad-induced events, we end up with a clean ON_SOURCE_LOADED/ON_SOURCE_UNLOADED
+    // sequence which only concerns the actual source.
+    this.updateSession();
```
protyposis: Shouldn't you initialize a new session when a new source is loaded?

src/ts/ConvivaAnalytics.ts
```diff
@@ -186,89 +218,73 @@ export class ConvivaAnalytics {
...
   };
 
   private onPlaybackStateChanged = (event?: any) => {
     this.debugLog('reportplaybackstate', event);
-    let playerState = Conviva.PlayerStateManager.PlayerState.UNKNOWN;
+    let playerState;
 
-    if ((!this.player.isPlaying() && !this.player.isPaused()) || this.player.hasEnded()) {
-      // Before playback has started, and after it is finished, we report the stopped state
-      playerState = Conviva.PlayerStateManager.PlayerState.STOPPED;
```
protyposis, line 225: Why don't you set the stopped state from here, but all other states you do?

src/ts/ConvivaAnalytics.ts
```diff
@@ -186,89 +218,73 @@ export class ConvivaAnalytics {
...
-      }
-
-      // Start a new session (also updates the playback state)
-      this.startSession(event);
-      // On calling play, playback is not immediately started, but the loading phase begins
-      this.playbackStarted = false;
-    } else {
-      // A normal play event happened, just update the playback state
-      this.onPlaybackStateChanged(event);
+    this.updateSession();
```
protyposis: Does it make sense to send the same unchanged static metadata every time the user presses "play"?

src/ts/ConvivaAnalytics.ts
```diff
@@ -186,89 +218,73 @@ export class ConvivaAnalytics {
...
+    if (!this.hasPlayingEvent) {
+      this.onPlaybackStateChanged();
     }
   };
 
+  private onPlaying = (event: any) => {
+    this.debugLog('playing', event);
+    this.updateSession();
+    this.playerStateManager.setPlayerState(Conviva.PlayerStateManager.PlayerState.PLAYING);
```
protyposis: Why do you set the player state directly here (and in the next few lines)? Wouldn't it be cleaner to hande the same state always from the same method?

src/ts/ConvivaAnalytics.ts
```diff
@@ -186,89 +218,73 @@ export class ConvivaAnalytics {
...
-      // Start a new session (also updates the playback state)
-      this.startSession(event);
-      // On calling play, playback is not immediately started, but the loading phase begins
-      this.playbackStarted = false;
-    } else {
-      // A normal play event happened, just update the playback state
-      this.onPlaybackStateChanged(event);
+    this.updateSession();
+    if (!this.hasPlayingEvent) {
+      this.onPlaybackStateChanged();
```
protyposis: Why `onPlaybackStateChanged()`, not `onPlaying()`?

src/ts/ConvivaAnalytics.ts
```diff
@@ -61,6 +64,7 @@ export class ConvivaAnalytics {
         + 'Please load the Conviva script (conviva-core-sdk.min.js) before Bitmovin\'s ConvivaAnalytics integration.');
       return; // Cancel initialization
     }
+    this.hasPlayingEvent = Boolean(player.EVENT.ON_PLAYING);
```
protyposis: Blocks usually group code that somehow belongs together. Does this state flag belong to the assertion above? Please insert a blank line and then add a comment about why this event state is tracked.

src/ts/ConvivaAnalytics.ts
```diff
@@ -130,51 +135,78 @@ export class ConvivaAnalytics {
...
-    if (assetId && assetTitle) {
-      assetName = `${assetId} ${assetTitle}`;
-    } else if (assetId && !assetTitle) {
-      assetName = assetId;
-    } else if (assetTitle && !assetId) {
-      assetName = assetTitle;
     } else {
-      assetName = 'Untitled (no source.title/source.contentId set)';
+      this.contentMetadata.assetName = 'unknown';
+      this.contentMetadata.streamUrl = 'unknown';
```
protyposis, line 155: Wouldn't it be cleaner to not report unknown data at all instead of reporting it, thus giving the system the impression that we know it, but then writing in the content that we don't know?

src/ts/ConvivaAnalytics.ts
```diff
@@ -130,51 +135,78 @@ export class ConvivaAnalytics {
...
-    this.sessionKey = this.client.createSession(contentMetadata);
+    this.sessionKey = this.client.createSession(this.contentMetadata);
 
     if (!this.isValidSession()) {
       // Something went wrong. With stable system interfaces, this should never happen.
       this.logger.consoleLog('Something went wrong, could not obtain session key',
         Conviva.SystemSettings.LogLevel.ERROR);
     }
 
+    this.playerStateManager.setPlayerState(Conviva.PlayerStateManager.PlayerState.STOPPED);
```
protyposis, line 167: What if autoplay is enabled? is the player still stopped?

protyposis (changes requested):
I have the impression that this PR was done in a hurry (a.k.a. sloppy) so I'm giving you some food for thought instead of just saying what's off ;) 

Please put a little more effort into code style.

Also please minimize the times metadata is sent. There should only be two times when metadata is sent to Conviva:
 1. When the session is created / the source is loaded
 2. When playback actually started and all missing metadata is known (with players with the ON_PLAYING event)
After that, this data stays static until a new session is created when a new source is loaded.

src/ts/ConvivaAnalytics.ts
offnertho: Because it would reach the playing state too early. When we don't have an ON_PLAYING event, we want to trigger the playing state in the first ON_TIME_CHANGED.

src/ts/ConvivaAnalytics.ts
offnertho: I thought about that too, and wanted to reduce side effects...

src/ts/ConvivaAnalytics.ts
offnertho: I was sticking to the documentation, which states that these fields are required. 'https://community.conviva.com/site/global/platforms/other_platforms/sdks/javascript_sdk/taskref/index.gsp#metadata_ref'

src/ts/ConvivaAnalytics.ts
offnertho: Yes. At the time this code executes it is. The state is then transitioned in ON_PLAYING (or ON_TIME_CHANGED).

src/ts/ConvivaAnalytics.ts
offnertho: Because I set the stopped state explicitly in the ON_PLAYBACK_FINISHED handler.

src/ts/ConvivaAnalytics.ts
offnertho: Yeah I could add a flag indicating that we already send the data, but did not want to do that.

src/ts/ConvivaAnalytics.ts
offnertho: 👍

src/ts/ConvivaAnalytics.ts
protyposis: I see but I think it's much cleaner if one and the same matter is always handled in one and the same place.

src/ts/ConvivaAnalytics.ts
protyposis: The flag is much cheaper than a multitude of unnecessary network requests, and it's a sign of a suboptimal archtitecture if unnecessary things are done.

src/ts/ConvivaAnalytics.ts
protyposis: Ok then the question is what required means... required from the beginning or required to be set sometime during the session? If the latter, then you can still set it `unknown` in `ON_SOURCE_LOADED` where it should actually be known.

protyposis (approved):
Seems legit 👍

PR #4, opened 2018-09-17
Fix issue that the viewerID sometimes is undefined

stonko1994:
## Description
With the latest conviva SDK version the viewerID sometimes was undefined. They changed when attributes should be updated.

_Please also do a quick test_

## Fix
Change session initialization and `contentMetadata` updates.

### Discussion
Please reply to my comments below.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/)
+and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [develop]
+
+### Added
+- Support for Conviva-SDK version 2.146.0.36444
+
+## [95c526]
+
+Support for Conviva-SDK version below 2.146.0.36444
+
+[95c526]: https://github.com/bitmovin/bitmovin-player-analytics-conviva/commit/95c526a7306cef98061f8f65e3dec3023df501af
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Change Log
+# Changelog
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
@@ -6,8 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
-### Added
-- Support for Conviva-SDK version 2.146.0.36444
+### Changed
+- Updated Conviva-SDK to 2.146.0.36444
 
 ## [95c526]
 
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # Bitmovin Player Conviva Analytics Integration
 
+## Compatibility
+**This version of the Conviva Analytics Integration works only with a Player Version 7.x.
+And the preferred version of the Coviva SDK is 2.146.0.36444.**
+
 ## Limitations
 This Conviva integration currently does not track pre-roll ads. The analytics session is created after pre-roll ads, before the main content is started. This is implemented according to Conviva's guidelines but is still considered a limitation as the `Exit Before Video Start` metrics cannot be measured. All other metrics are working as expected.
 
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 # Bitmovin Player Conviva Analytics Integration
 
 ## Compatibility
-**This version of the Conviva Analytics Integration works only with a Player Version 7.x.
-And the preferred version of the Coviva SDK is 2.146.0.36444.**
+**This version of the Conviva Analytics Integration works only with Player Version 7.x.
+The recommended version of the Conviva SDK is 2.146.0.36444.**
 
 ## Limitations
 This Conviva integration currently does not track pre-roll ads. The analytics session is created after pre-roll ads, before the main content is started. This is implemented according to Conviva's guidelines but is still considered a limitation as the `Exit Before Video Start` metrics cannot be measured. All other metrics are working as expected.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 ## Compatibility
 **This version of the Conviva Analytics Integration works only with Player Version 7.x.
-The recommended version of the Conviva SDK is 2.146.0.36444.**
+The recommended version of the Conviva SDK is 2.146.0.36444.** See [CHANGELOG](CHANGELOG.md) for details.
 
 ## Limitations
 This Conviva integration currently does not track pre-roll ads. The analytics session is created after pre-roll ads, before the main content is started. This is implemented according to Conviva's guidelines but is still considered a limitation as the `Exit Before Video Start` metrics cannot be measured. All other metrics are working as expected.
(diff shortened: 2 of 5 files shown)


src/ts/Conviva.d.ts
```diff
@@ -75,6 +75,9 @@ declare namespace Conviva {
     applicationName: string;
     assetName: string;
     custom: {};
+    /**
+     * @deprecated in at SDK 2.146.0.36444
```
stonko1994: Should we keep this? 

Since we don't have to dependency to a specific conviva version i'm not sure.
Maybe we should create this dependency? Maybe just with a comment in the Readme, or a version check somewhere in the code and display a warning?

src/ts/ConvivaAnalytics.ts
```diff
@@ -174,35 +185,42 @@ export class ConvivaAnalytics {
...
+   * Update contentMetadata which must be present before first video frame
+   */
+  private buildContentMetadata() {
     let source = this.player.getConfig().source;
 
-    if (this.contentMetadata.assetName !== this.getAssetName(source)) {
-      this.contentMetadata.assetName = this.getAssetName(source);
-    }
-
+    this.contentMetadata.applicationName = this.config.applicationName || 'Unknown (no config.applicationName set)';
```
stonko1994, line 190: Should we make `config.applicationName` mandatory? Or even not available outside because this will result in the `Player Name` field on pulse.conviva

src/ts/Conviva.d.ts
protyposis: If unused, then please remove

protyposis: Yes we need to introduce some kind of version management. As a first step, please add to the readme (somewhere on the top where it can be seen easily), which version(s) of the Conviva SDK this has been developed/tested for.

src/ts/ConvivaAnalytics.ts
protyposis: > Or even not available outside because this will result in the `Player Name` field on pulse.conviva

What do you mean by that? Removing the "unknown" fallback string and only setting it if available?

src/ts/ConvivaAnalytics.ts
```diff
@@ -174,35 +185,42 @@ export class ConvivaAnalytics {
...
+  }
+
+  /**
+   * Update contentMetadata which are allowed during the session
+   */
+  private updateSession() {
+    if (!this.isValidSession()) {
+      return;
+    }
+    this.contentMetadata.streamUrl = this.getUrlFromSource(this.player.getConfig().source); // TODO: mid Meta data available
```
protyposis: What does the TODO mean?

stonko1994: Was a note for myself. Forgot to remove the TODO.

src/ts/ConvivaAnalytics.ts
stonko1994: I mean that we could consider making the `applicationName` mandatory in the config. So that we can remove the fallback string. Since `applicationName` will result in the `Player Name` field on pulse i think this should be set always to see were the session is coming from. But it's just my opinion and i'm not 100% sure about that.

src/ts/Conviva.d.ts
stonko1994: Should we then also add a changelog so that we can say `Version x.y is compatible with Conviva SDK version a.b`?

src/ts/ConvivaAnalytics.ts
protyposis: The fallback is only there to make it obvious that the `applicationName` has not been set (and maybe because it was mandatory in the past) - so this makes it kind of mandatory: if you don't want the unknown warning, set the app name.

src/html/test.html
```diff
@@ -56,19 +56,22 @@
...
-    source: {
-      title: 'Art of Motion',
-      description: 'What is this event... Parcour?',
-      dash: '//bitmovin-a.akamaihd.net/content/MI201109210084_1/mpds/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.mpd',
-      hls: '//bitmovin-a.akamaihd.net/content/MI201109210084_1/m3u8s/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.m3u8',
-      progressive: '//bitmovin-a.akamaihd.net/content/MI201109210084_1/MI201109210084_mpeg-4_hd_high_1080p25_10mbits.mp4',
-      poster: '//bitmovin-a.akamaihd.net/content/art-of-motion_drm/art-of-motion_poster.jpg',
-      viewerId: 'uniqueViewerIdThatOverridesTheConvivaAnalyticsConfig', // Conviva viewerId
-      contentId: 'uniqueContentId', // Conviva contentId (will be prefixed to the title)
+    source: source,
```
offnertho, line 71: should we set the source in the configuration? 

With v8 this is not valid anymore.

src/ts/Conviva.d.ts
protyposis: Since we don't have versioning yet you can add a changelog with git commit hashes in the meantime. That would be nice.

offnertho (commented):
Looks good. Will have to test it.

src/html/test.html
stonko1994: For now yes.

This is only a fix for the v7 player-conviva-integration. The v8 upgrade will happen later.

offnertho (approved):
Seems to work 👍

offnertho (approved):

stonko1994:
@protyposis please have another look at the added Changelog.
EDIT: and the Readme.

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
+# Change Log
```
protyposis: `Changelog` should be written as a single word

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/)
+and this project adheres to [Semantic Versioning](http://semver.org/).
+
+## [develop]
+
+### Added
+- Support for Conviva-SDK version 2.146.0.36444
```
protyposis: Does this mean that this version is supported additionally to the previous ones, or is only this version supported?

README.md
```diff
@@ -1,5 +1,9 @@
 # Bitmovin Player Conviva Analytics Integration
 
+## Compatibility
+**This version of the Conviva Analytics Integration works only with a Player Version 7.x.
+And the preferred version of the Coviva SDK is 2.146.0.36444.**
```
protyposis: Remove the `And`. Just `The recommended version of the...`

src/ts/ConvivaAnalytics.ts
```diff
@@ -129,6 +128,9 @@ export class ConvivaAnalytics {
   }
 
   private getUrlFromSource(source: any): string {
+    if (!source) {
```
protyposis: How can this case ever happen? If there is no source, there shouldn't be a session because there's nothing to track?

protyposis (changes requested):

CHANGELOG.md
stonko1994: No this version supports only this Conviva-SDK version.

src/ts/ConvivaAnalytics.ts
stonko1994: You are right, it's not possible.

CHANGELOG.md
protyposis: Then this should be in the `Changed` section as it does not *add* support for this version (additionally) but rather changes the version requirement of the SDK: `Updated Conviva SDK to v2.1....`

README.md
```diff
@@ -1,5 +1,9 @@
 # Bitmovin Player Conviva Analytics Integration
 
+## Compatibility
+**This version of the Conviva Analytics Integration works only with Player Version 7.x.
+The recommended version of the Conviva SDK is 2.146.0.36444.**
```
protyposis: I forgot in the previous review: please add a link to the changelog. Just write `See [CHANGELOG](CHANGELOG.md) for details.`.

protyposis (changes requested):

protyposis (approved):

PR #5, opened 2018-10-15
Updates according to conviva validation feedback

stonko1994:
## Description
Changes requested by the verification process of conviva:
- Initialize session only when the user requests playback (when the user presses play or autoplay content) - Was pointed out by a wrong VST (Video Startup Time)
- Bind lifecycle from the `PlayerStateManager` to the lifecycle of the `convivaSession`
- Track VSF (Video Startup Failures)
- Do not track events from the ad player

Suggestions from conviva:
- Track `autoplay` and `preload` config since they are affecting VST
- Add customTags for:
  - appVersion
  - contentId
  - playerVendor
  - playerVersion (also tracked automatic in `PlayerStateManager`)

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -196,7 +196,15 @@ export class ConvivaAnalytics {
     this.contentMetadata.streamType =
       this.player.isLive() ? Conviva.ContentMetadata.StreamType.LIVE : Conviva.ContentMetadata.StreamType.VOD;
 
-    this.contentMetadata.custom = this.config.customTags;
+    this.contentMetadata.custom = {
+      'playerType': this.player.getPlayerType(),
+      'streamType': this.player.getStreamType(),
+      'vrContentType': this.player.getVRStatus().contentType,
+      // Autoplay and preload are important options for the Video Startup Time so we track it as custom tags
+      'autoplay': PlayerConfigHelper.getAutoplayConfig(this.player) + '',
+      'preload': PlayerConfigHelper.getPreloadConfig(this.player) + '',
+      ...this.config.customTags,
+    };
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/ts/ConvivaAnalytics.ts
```diff
@@ -427,8 +449,6 @@ export class ConvivaAnalytics {
   private registerPlayerEvents(): void {
     let player = this.player;
     let playerEvents = this.playerEvents;
-    playerEvents.add(player.EVENT.ON_SOURCE_LOADED, this.onSourceLoaded);
```
protyposis, line 430: Is it really ok to remove these? Don't we have to track the time between loading the player and starting playback as well? What if we have a preroll ad, is the metadata still correctly set?

protyposis (approved):
Code looks fine, a few unclear points though.

src/ts/ConvivaAnalytics.ts
stonko1994: No we don't need to track time between loading and player start playback. Only interesting value here is VST (Video Startup Time) Time between User initializes play and the player starts playback. Therefore it was suggested to track if preload / autoplay is enabled because this is effecting the VST the most.

protyposis: 👍

src/ts/ConvivaAnalytics.ts
```diff
@@ -243,32 +250,13 @@ export class ConvivaAnalytics {
     return this.sessionKey !== Conviva.Client.NO_SESSION_KEY;
   }
 
-  private onSourceLoaded = () => {
-    if (this.isAd) {
-      // Ignore ON_SOURCE_LOADED events during ad playback, because that's just an ad being temporarily loaded
-      // instead of the actual source.
+  private onPlaybackStateChanged = (event?: any) => {
+    this.debugLog('reportplaybackstate', event);
+    if (event && event.issuer === 'IMA') {
```
offnertho, line 255: is this check really save? 

I guess this check should prevent the playback state update from Ad Events right? 

So what if the issuer is not `IMA` for example when the Ad does not get handled by the IMA SDK?

src/ts/ConvivaAnalytics.ts
```diff
@@ -488,12 +510,63 @@ export class ConvivaAnalytics {
...
+  /**
+   * Extract preload config from player
+   *
+   * The preload config can be set individual for mobile or desktop.
+   * Default value is true for VOD and false for live streams. If the value is not set for current platform
+   * the default value will be used over the value for the other platform.
+   *
+   * @param player: Player
+   */
+  public static getPreloadConfig(player: Player): boolean {
```
offnertho, line 549: The preload can also be set on root level: 
```
playerConfig.adaptation.preload
```

offnertho (changes requested):
Looks good. 

Just some small remarks.

src/ts/ConvivaAnalytics.ts
stonko1994: No it isn't. It's just a naive approach to silence playback state changes during an ad. I'm not totally sure why those play events are even triggered. But at least for the case we play an IMA ad it is working.

If this check is not working it's not the end of the world because conviva is suppress all playback state change events between `adStarted` and `adEnded` by default. But anyway they should not be send.

src/ts/ConvivaAnalytics.ts
stonko1994: Thanks. Didn't know that.

offnertho (approved):

src/ts/ConvivaAnalytics.ts
protyposis: That wasn't necessary. The player setup expands a root level setting into the `mobile` and `desktop` subsettings.

PR #6, opened 2018-10-18
Minor improvements regarding conviva validation

stonko1994:
## Description
There are some more requests to change from conviva.

- Do not track adFinished events if an adError occured but no ad ever started.
- Do not track playing events if it transits into stalled within nearly 0 ms
- Send all available content metadata attributes right on session creation.

## Fix
Fix for the second issue is to check if the buffer is 0 in the playing event. If yes we do not report this event. As discussed with @bitmovin-kenny this playing event is natural behavior of the video-element.

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -299,12 +299,6 @@ export class ConvivaAnalytics {
   };
 
   private onPlaying = (event: any) => {
-    // When seeking over the buffer the player emits an onPlaying event followed by an onStallStarted within 0 ms.
-    // To suppress this play state transition we check the buffer if the video element is really ready to start playing.
-    if (this.player.getVideoBufferLength() === 0) {
-      this.debugLog('no buffer - skipped playing event', event);
-      return;
-    }
     this.playbackStarted = true;
     this.debugLog('playing', event);
     this.updateSession();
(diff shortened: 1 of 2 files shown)


src/ts/ConvivaAnalytics.ts
```diff
@@ -205,16 +204,24 @@ export class ConvivaAnalytics {
...
     };
+
+    // also include dynamic content metadata at initial creation
+    this.buildDynamicContentMetadata();
   }
 
   /**
    * Update contentMetadata which are allowed during the session
    */
+  private buildDynamicContentMetadata() {
```
offnertho, line 215: why is the url dynamic?

offnertho, line 215: why is this named `dynamicContentMetadata`.

src/ts/ConvivaAnalytics.ts
```diff
@@ -292,6 +299,12 @@ export class ConvivaAnalytics {
   };
 
   private onPlaying = (event: any) => {
+    // When seeking over the buffer the player emits an onPlaying event followed by an onStallStarted within 0 ms.
+    // To suppress this play state transition we check the buffer if the video element is really ready to start playing.
+    if (this.player.getVideoBufferLength() === 0) {
```
offnertho: what if it is an audio only stream?

offnertho (changes requested):
Pls take care of audio only streams. 

This is all for `v7`, how do we plan on supporting `v8`.

src/ts/ConvivaAnalytics.ts
stonko1994: Good point. 

So is for an audio only stream the `getVideoBufferLength` always 0 (and vice versa)?

stonko1994: Good point. 

So is for an audio only stream the `getVideoBufferLength` always 0 (and vice versa)?

src/ts/ConvivaAnalytics.ts
stonko1994: hm. was the best wording i found for content metadata that can be changed over time.

stonko1994: It's defined by conviva what can be changed during playback. And i just wanted to support updating (conviva) supported attributes.

src/ts/ConvivaAnalytics.ts
protyposis: Since this is a bug in the player that we need to fix, this check should probably be removed?

stonko1994: Yes, I'm waiting for a response of conviva but it will be removed anyway.

stonko1994:
I got the recommendation from conviva that we should remove seek event tracking for now until the player-bug about the playing event is fixed. 

They know that removing seeking won't change the playing event but it will sanitize their metric calculations because there will be no trigger for certain logic which in this case isn’t reliable.

Please review again.

protyposis (approved):

offnertho (approved):

PR #8, opened 2018-11-08
Fix an issue that a post-roll ad was tracked as a new conviva session

stonko1994:
## Description
We are creating a new session when an `onPlaying` gets triggered. This will be also triggered when an ad starts. We suppress this `playing` event with the `isAd` flag but the `adStarted` event comes after a `playing` event of an ad. So we created a new session for a post-roll ad.

## Fix
Set the `isAd` flag within the `adBreakStarted` / `adBreakFinished` events.

Minor improvement:
- only log events when we actually report them.
- close session on player destroy

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -424,6 +424,10 @@ export class ConvivaAnalytics {
     this.endSession(event);
   };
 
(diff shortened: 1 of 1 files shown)


offnertho (approved):
Looks good.

PR #9, opened 2018-12-13
Feature/player v8

fabiandev:
This PR includes the following breaking changes and therefore needs a major version release:
- Update to Player v8
- Switch to webpack
- Improve typings
- Make package publishable


CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
+## Changed
+- Update to Bitmovin Player v8
+- Switch to Webpack and yarn
```
protyposis: Which team is responsible for this project now? If it's still the player team, please revert to `npm`. We are not using `yarn` as there is no functional advantage over recent `npm` versions.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [develop]
 
+## Changed
+- Update to Bitmovin Player v8
```
protyposis, line 10: We need to split a `support/v7` branch before updating `develop` to v8

example/index.html
```diff
@@ -0,0 +1,23 @@
...
+  </head>
+  <body>
+  <div class="container">
+    <div class="content">
+      <div class="player-wrapper">
+        <div id="player"></div>
+      </div>
+    </div>
+  </div>
+  <script src="./script.js"></script>
```
protyposis: An example should be written with simplicity in mind and as easy to understand as possible. Extracting the JS code into a separate file introduces an unnecessary indirection for people trying to understand the code, especially inexperienced ones. I know this is a cleaner "architecture" but that's not the main focus of an example.

src/ts/Conviva.ts
```diff
@@ -0,0 +1,205 @@
+export declare namespace Conviva {
```
protyposis: Declarations should be in a separate file

src/ts/main.ts
```diff
@@ -1,8 +1,9 @@
...
 // Export Conviva Analytics to global namespace
-let w = (window as any);
-w.bitmovin = w.bitmovin || {};
-w.bitmovin.player = w.bitmovin.player || {};
-w.bitmovin.player.analytics = w.bitmovin.player.analytics || {};
-w.bitmovin.player.analytics.ConvivaAnalytics = ConvivaAnalytics;
+window.bitmovin = window.bitmovin || {};
+window.bitmovin.player = window.bitmovin.player || {};
+window.bitmovin.player.analytics = window.bitmovin.player.analytics || {};
+window.bitmovin.player.analytics.ConvivaAnalytics = ConvivaAnalytics;
```
protyposis: Replace this outdated manual "hack" with a correct Webpack UMD export configuration.

src/types/global.d.ts
```diff
@@ -0,0 +1,15 @@
+import BitmovinPlayer from 'bitmovin-player';
+import ConvivaAnalytics from '../ts/ConvivaAnalytics';
+
+declare global {
+  interface Window {
+    bitmovin: {
+      player?: {
+        Player?: typeof BitmovinPlayer,
+        analytics?: {
+          ConvivaAnalytics?: typeof ConvivaAnalytics,
```
protyposis: Global exports are undesired and only a fallback if the JS file is loaded directly into the webpage without a module loader, in which case it will be JS (not TS), so type declarations for the global export are unnecessary and discouraged.

protyposis (changes requested):
This is a high level review without functional testing. An estimated 95% of the changes in here are unrelated to the v8 upgrade and make it very unnecessarily difficult to track the actually required changes for v8. Of these 95% the majority are a question of subjective personal taste and technically not necessary.

Also by deleting "old" files and inserting "new" files, instead of moving and changing them, you've made it impossible to track changes. Also moving huge blocks of code between files, e.g. the type declarations, made it impossible to track the allegedly added improvements.

src/ts/ConvivaAnalytics.ts
```diff
@@ -71,20 +87,14 @@ export class ConvivaAnalytics {
...
     }
 
-    // player versions <=7.2 did not have a ON_PLAYING event
-    // we track this change to correctly transition to the playing state
-    this.hasPlayingEvent = Boolean(player.EVENT.ON_PLAYING);
     this.sessionDataPopulated = false;
 
-    // Assert that this class is instantiated before player.setup() is called.
-    // When instantiated later, we cannot detect startup error events because they are fired during setup.
-    if (player.isReady()) {
```
stonko1994, line 81: Please check if there is already a source loaded instead.

```ts
if (!player.getSource()) {
   console.error('ConvivaAnalytics must be instantiated before calling player.load()');
  return; // Cancel initialization
}
```

src/ts/ConvivaAnalytics.ts
```diff
@@ -405,18 +429,18 @@ export class ConvivaAnalytics {
...
 
-  private onError = (event: any) => {
+  private onError = (event: ErrorEvent) => {
     if (!this.isValidSession()) {
       // initialize Session if not yet initialized to capture Video Start Failures
       this.initializeSession();
     }
 
-    this.client.reportError(this.sessionKey, String(event.code) + ' ' + event.message,
+    this.client.reportError(this.sessionKey, `${String(event.code)} ${event.name}: `,
```
stonko1994: change to `${String(event.code)} ${event.name}`. Remove the colon at the end.

src/ts/Html5Http.ts
```diff
@@ -20,9 +14,45 @@ export class Html5Http implements Conviva.HttpInterface {
     return this.makeRequestStandard.apply(this, arguments);
   }
 
+  //   this.makeRequestIE89 = function (httpMethod, url, data, contentType, timeoutMs, callback) {
```
stonko1994: please remove those comments.

stonko1994 (changes requested):
I would have preferred if there would only be v8 update related changes.

However i did a quick test and we will have to wait for v8.2.0-rc1 to make a detailed test run.

src/ts/ConvivaAnalytics.ts
fabiandev: Should be `if (player.getSource())` w/o `!`, right?

fabiandev:
For only Player v8 related changes, please look at the [diff](https://github.com/bitmovin/bitmovin-player-analytics-conviva/pull/9/files#diff-44a026b2fab5c7ae064ab659237f4333) of [`src/ts/ConvivaAnalytics.ts`](https://github.com/bitmovin/bitmovin-player-analytics-conviva/blob/feature/player-v8/src/ts/ConvivaAnalytics.ts). This file has not been moved and mostly includes changes related to the player upgrade, including added types. The other changes are build related, due to outdated packages and vulnerabilities.

src/ts/Conviva.ts
fabiandev: What do you mean by that, since this is a separate file?

protyposis: Sorry I thought you moved it into `ConvivaAnalytics.ts` because I assumed it would be a simple file rename otherwise.

fabiandev: No worries, will revert it to `.d.ts`, though.

src/ts/ConvivaAnalytics.ts
stonko1994: true sry.

src/ts/Conviva.d.ts
```diff
@@ -169,12 +161,12 @@ declare namespace Conviva {
 
   class SystemFactory {
     constructor(systemInterface: SystemInterface, systemSettings: SystemSettings);
-    release(): void;
+    public release(): void;
   }
 
   class SystemInterface {
     constructor(timeInterface: TimeInterface, timerInterface: TimerInterface, httpInterface: HttpInterface,
-        storageInterface: StorageInterface, metadataInterface: MetadataInterface, loggingInterface: LoggingInterface);
```
stonko1994, line 177: please put every property in a single line

src/ts/Conviva.d.ts
```diff
@@ -39,28 +39,28 @@ declare namespace Conviva {
 
   class Client {
     constructor(settings: ClientSettings, systemFactory: SystemFactory);
-    adEnd(sessionKey: number): void;
-    adStart(sessionKey: number, adStream: Client.AdStream, adPlayer: Client.AdPlayer,
+    public adEnd(sessionKey: number): void;
+    public adStart(sessionKey: number, adStream: Client.AdStream, adPlayer: Client.AdPlayer,
```
stonko1994: please put every property in a single line

src/ts/ConvivaAnalytics.ts
```diff
@@ -117,13 +131,47 @@ export class ConvivaAnalytics {
...
+    if (!this.isValidSession()) {
+      this.logger.consoleLog('cannot send playback event, no active monitoring session',
+        Conviva.SystemSettings.LogLevel.WARNING);
+      return;
+    }
+
+    this.client.sendCustomEvent(this.sessionKey, eventName, eventAttributes);
+  }
+
+  public release(event?: PlayerEventBase): void {
```
stonko1994: I think that `release()` shouldn't have a parameter event. This is just confusing when a user needs to call this. Event if it's optional.

src/ts/ConvivaAnalytics.ts
```diff
@@ -370,52 +400,59 @@ export class ConvivaAnalytics {
         break;
     }
 
+    // TODO: check if also should not report for post-roll (currently doesn't)
```
stonko1994: It is valid as it is now.

src/ts/ConvivaAnalytics.ts
```diff
@@ -370,52 +400,59 @@ export class ConvivaAnalytics {
...
+    this.debugLog('adbreakstart', event);
     this.client.adStart(this.sessionKey, Conviva.Client.AdStream.SEPARATE, Conviva.Client.AdPlayer.CONTENT, adPosition);
     this.onPlaybackStateChanged();
   };
 
-  private onAdSkipped = (event: any) => {
-    this.onCustomEvent(event);
-    this.onAdFinished(event);
-  };
+  private getAdBreakPosition(adBreak: AdBreak) {
```
stonko1994: I would suggest do make this to something like `mapAdPosition` and return the conviva enum value. So we don't have to compare the strings afterwards again.

src/ts/Conviva.d.ts
protyposis: Including a trailing comma after the last parameter, i.e.

```ts
public adStart(
  sessionKey: number, 
  adStream: Client.AdStream, 
  adPlayer: Client.AdPlayer,
  adPosition: Client.AdPosition,
): void;
```

src/ts/Html5Timer.ts
```diff
@@ -1,32 +1,19 @@
...
 import TimerCancelFunction = Conviva.TimerCancelFunction;
 import TimerAction = Conviva.TimerAction;
 
 export class Html5Timer implements Conviva.TimerInterface {
 
-  createTimer(timerAction: TimerAction, intervalMs: number, actionName?: string | null): TimerCancelFunction {
+  public createTimer(timerAction: TimerAction, intervalMs: number, actionName?: string | null): TimerCancelFunction {
     let timerId = setInterval(timerAction, intervalMs);
-    let cancelTimerFunc = (function() {
+    return (function () {
```
protyposis: No whitespace in `function()`

tsconfig.json
```diff
@@ -1,11 +1,15 @@
...
-    "declaration": true,
-    "outDir": "dist/js/framework"
-  }
-}
\ No newline at end of file
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "lib": ["es2015", "dom"],
+    "sourceMap": true,
+    "noImplicitAny": false,
```
protyposis: Why did you change this to `false`?

tsconfig.json
```diff
@@ -1,11 +1,15 @@
...
+    "removeComments": true
```
protyposis: Why remove the comments? The minified Webpack build will not contain them anyway, and the transpiled modules can contain them for better understandability if anyone looks into them (when using modules from the NPM package).

src/ts/Conviva.d.ts
```diff
@@ -39,28 +39,28 @@ declare namespace Conviva {
 
   class Client {
     constructor(settings: ClientSettings, systemFactory: SystemFactory);
-    adEnd(sessionKey: number): void;
-    adStart(sessionKey: number, adStream: Client.AdStream, adPlayer: Client.AdPlayer,
+    public adEnd(sessionKey: number): void;
```
protyposis, line 43: The changelog says that the type declarations were improved, but the only difference I can see here are unnecessary `public` modifiers, so semantically there is no difference and the changelog entry should be removed - or is there anything I have missed?

src/ts/Html5Http.ts
```diff
@@ -1,28 +1,20 @@
-///<reference path="../ts/Conviva.d.ts"/>
-/*! (C) 2016 Conviva, Inc. All rights reserved. Confidential and proprietary. */
```
protyposis, line 2: All TS files that contain(ed) this header are files from the Conviva SDK, that were simply converted to TS but otherwise untouched, so please keep all those headers within the files. Don't make their code seem like ours.

src/ts/Html5Http.ts
```diff
@@ -1,28 +1,20 @@
...
 export class Html5Http implements Conviva.HttpInterface {
 
-  makeRequest(httpMethod: 'GET' | 'POST', url: string, data: string | null, contentType: string | null,
+  public makeRequest(httpMethod: 'GET' | 'POST', url: string, data: string | null, contentType: string | null,
               timeoutMs: number, callback: HttpRequestCallback | null): HttpRequestCancelFunction {
-    // XDomainRequest only exists in IE, and is IE8-IE9's way of making CORS requests.
-    // It is present in IE10 but won't work right.
-    // if (typeof XDomainRequest !== 'undefined' && navigator.userAgent.indexOf('MSIE 10') === -1) {
-    // 	return this.makeRequestIE89.apply(this, arguments);
-    // }
```
protyposis, line 19: This comment, as the others below, are part of the Conviva SDK source files so please keep it in place.

protyposis (changes requested):

src/ts/Conviva.d.ts
fabiandev: The typings of `ConvivaAnalytics.ts` have been improved, also because of player v8. I did not touch the ambient typings of the Conviva API. E.g., `Player` was typed as `any` before.

src/ts/Html5Http.ts
fabiandev: @dweinber @stonko1994 Please let me know how to proceed on this one, since we discussed that before.

src/ts/Html5Http.ts
fabiandev: @stonko1994 https://github.com/bitmovin/bitmovin-player-analytics-conviva/pull/9#discussion_r241367837

src/ts/Html5Http.ts
stonko1994: we decided together with @dweinber to remove those comments from all files.

src/ts/ConvivaAnalytics.ts
```diff
@@ -404,9 +404,8 @@ export class ConvivaAnalytics {
         break;
     }
 
-    // TODO: check if also should not report for post-roll (currently doesn't)
     if (!this.isValidSession()) {
-      // Don't report without a valid session (e.g. in case of a post-roll ad)
+      // Don't report without a valid session (e.g. in case of a pre-roll ad)
```
stonko1994: please leave the comment as it was. In case of a post-roll ad the session already got shut down. So the old comment fits better.

fabiandev: But also there is not yet a session in case of a pre-roll ad. I did adjust the comment accordingly.

protyposis (approved):
I'm making an exception for this PR and approve it, but please remember for future PRs that they should be kept as small as possible (this one can be split into multiple PRs like "Change build to webpack", "Update type definitions", "Update to player 8", etc) and huge PRs like this will be rejected.

stonko1994 (approved):

PR #10, opened 2019-01-16
Add version of current integration as custom tag

stonko1994:
## Description
This will add the current version of the integration as custom tag which is tracked to conviva.

## TODO:

- [x] forward port


protyposis (approved):
Nice!

PR #11, opened 2019-01-17
Add possibility to track custom deficiency events

stonko1994:
## Description
Since we don't want to add a timeout in certain use-cases where the player stalls for a long time we now add a method where the user of the integration can decide if they want to track a deficiency to conviva.

_In the future there will be more methods to track custom session events._

## TODO

- [x] forward port

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,6 @@
 **This version of the Conviva Analytics Integration works only with Player Version 7.x.
 The recommended version of the Conviva SDK is 2.146.0.36444.** See [CHANGELOG](CHANGELOG.md) for details.
 
-## Limitations
-This Conviva integration currently does not track pre-roll ads. The analytics session is created after pre-roll ads, before the main content is started. This is implemented according to Conviva's guidelines but is still considered a limitation as the `Exit Before Video Start` metrics cannot be measured. All other metrics are working as expected.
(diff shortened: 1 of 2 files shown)


dweinber (approved):
LGTM 👍

README.md
```diff
@@ -4,9 +4,6 @@
 **This version of the Conviva Analytics Integration works only with Player Version 7.x.
 The recommended version of the Conviva SDK is 2.146.0.36444.** See [CHANGELOG](CHANGELOG.md) for details.
 
-## Limitations
-This Conviva integration currently does not track pre-roll ads. The analytics session is created after pre-roll ads, before the main content is started. This is implemented according to Conviva's guidelines but is still considered a limitation as the `Exit Before Video Start` metrics cannot be measured. All other metrics are working as expected.
```
protyposis, line 8: How did you implement this?

protyposis (approved):

README.md
stonko1994: Some time ago we changed the way how we initialize the session. The session is initialized `onPlay` and this is triggered also in case of pre-roll ads. So this was no longer a limitation.

protyposis: Nice!

PR #13, opened 2019-01-24
Add version of current integration as custom tag

stonko1994:
## Description
Forward port of #10 with changing from `gulp` to `webpack`

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -1,5 +1,5 @@
+const npmPackage = require('./package.json');
 const path = require('path');
-const CopyPlugin = require('webpack-copy-plugin');
 
 module.exports = {
   entry: './src/ts/index.ts',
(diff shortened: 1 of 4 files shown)


webpack.common.js
```diff
@@ -9,6 +9,15 @@ module.exports = {
         test: /\.tsx?$/,
         use: 'ts-loader',
         exclude: /node_modules/
+      },
+      {
+        test: /\.tsx?$/,
+        loader: 'string-replace-loader',
+        options: {
+          search: '{{VERSION}}',
+          replace: npmPackage.version,
```
offnertho, line 18: we could probably also inline this: `require("./package.json").version`

offnertho (approved):
Looks good.

webpack.common.js
stonko1994: We could but I prefer as it is. (Have imports / requires on top of files)

PR #15, opened 2019-01-24
External session handling

stonko1994:
## Description
This PR adds two public methods which allows to manage the session externally.

## Discussion
Should we add those warnings from the comments to the Readme?


dweinber (approved):
LGTM

PR #16, opened 2019-01-29
Add ability to manage content metadata attribute externally

stonko1994:
## Description
This will add the possibility to override content metadata attributes externally.

### DIFF
--- a/src/ts/ContentMetadataBuilder.ts
+++ b/src/ts/ContentMetadataBuilder.ts
@@ -1,0 +1,128 @@
+export interface ModifiableMetadata {
+  // Static Attributes
+  viewerId?: string;
+  streamType?: Conviva.ContentMetadata.StreamType;
+  applicationName?: string;
+  custom?: CustomContentMetadata;
+  duration?: number;
+
+  // Dynamic
+  encodedFrameRate?: number;
+  defaultResource?: string;
+  streamUrl?: string;
+}
+
+interface Metadata extends ModifiableMetadata {
+  // Can only be set once
+  assetName?: string;
+}
+
+interface CustomContentMetadata {
+  [key: string]: string;
+}
+
+export class ContentMetadataBuilder {
+
+  private readonly logger: Conviva.LoggingInterface;
+  private readonly contentMetadata: Conviva.ContentMetadata;
+
+  // intern metadata fields to enable merging / overriding
+  private metadataOverrides: Metadata = {};
+  private metadata: Metadata = {};
+  private playbackStarted: boolean = false;
+
+  constructor(logger: Conviva.LoggingInterface) {
+    this.logger = logger;
+    this.contentMetadata = new Conviva.ContentMetadata();
+  }
+
+  /**
+   * This method is used for custom content metadata updates during / before a session.
+   * @param newValue
+   */
+  setOverrides(newValue: ModifiableMetadata) {
+    if (this.playbackStarted) {
+      this.logger.consoleLog(
+        '[ Conviva Analytics ] Playback has started. Only some metadata attributes will be updated',
+        Conviva.SystemSettings.LogLevel.WARNING,
+      );
+    }
+
+    this.metadataOverrides = newValue;
+  }
+
+  // Is needed to determine which attributes are allowed to update
+  setPlaybackStarted(value: boolean) {
+    this.playbackStarted = value;
+  }
+
+  build(): Conviva.ContentMetadata {
+    if (!this.playbackStarted) {
+      // Asset name is only allowed to be set once
+      if (!this.contentMetadata.assetName) {
+        this.contentMetadata.assetName = this.metadataOverrides.assetName || this.metadata.assetName;
+      }
+
+      this.contentMetadata.viewerId = this.viewerId;
+      this.contentMetadata.streamType = this.metadataOverrides.streamType || this.metadata.streamType;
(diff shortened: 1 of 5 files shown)


src/ts/ContentMetadataBuilder.ts
```diff
@@ -0,0 +1,130 @@
...
+interface CustomContentMetadata {
+  [key: string]: string;
+}
+
+export class ContentMetadataBuilder {
+
+  private readonly logger: Conviva.LoggingInterface;
+  private readonly contentMetadata: Conviva.ContentMetadata;
+
+  // intern metadata fields to enable merging / overriding
```
bitmovin-kenny:
```suggestion
  // internal metadata fields to enable merging / overriding
```

src/ts/ContentMetadataBuilder.ts
```diff
@@ -0,0 +1,130 @@
...
+      this.logger.consoleLog(
+        '[ Conviva Analytics ] Playback has started. Only some metadata attributes will be updated',
+        Conviva.SystemSettings.LogLevel.WARNING,
+      );
+    }
+
+    this.metadataOverrides = newValue;
+  }
+
+  // Is needed to determine which attributes are allowed to update
```
bitmovin-kenny: this comment looks out of place

src/ts/ConvivaAnalytics.ts
```diff
@@ -144,19 +123,18 @@ export class ConvivaAnalytics {
...
+    if (this.isValidSession()) {
+      this.logger.consoleLog('There is already a session running.', Conviva.SystemSettings.LogLevel.WARNING);
+      return;
     }
 
-    if (assetName) {
-      this.metadataOverrides.assetName = assetName;
+    // This could be called before source loaded.
+    // Without setting the asset name on the content metadata when initializing the session the SDK will throw errors.
+    if (!this.player.getSource() && !this.contentMetadataBuilder.assetName) {
```
bitmovin-kenny, line 137: shouldn't this be part of `isValidSession` ?

src/ts/ConvivaAnalytics.ts
```diff
@@ -144,19 +123,18 @@ export class ConvivaAnalytics {
...
+  public initializeSession(): void {
+    if (this.isValidSession()) {
+      this.logger.consoleLog('There is already a session running.', Conviva.SystemSettings.LogLevel.WARNING);
+      return;
     }
 
-    if (assetName) {
-      this.metadataOverrides.assetName = assetName;
+    // This could be called before source loaded.
+    // Without setting the asset name on the content metadata when initializing the session the SDK will throw errors.
```
bitmovin-kenny: `the session the SDK will throw errors` something is wrong here

src/ts/ConvivaAnalytics.ts
```diff
@@ -204,6 +182,31 @@ export class ConvivaAnalytics {
...
   }
 
+  /**
+   * Will update the contentMetadata which are tracked with conviva.
+   *
+   * If there is an active session only permitted values will be updated and propagated immediately.
+   * If there is no active session the values will set on session creation.
+   *
+   * Attributes set via this method will override automatic tracked once.
+   * @param metadata
```
bitmovin-kenny: missing jsDoc

src/ts/ConvivaAnalytics.ts
```diff
@@ -204,6 +182,31 @@ export class ConvivaAnalytics {
...
+   * @see ContentMetadataBuilder for more information about permitted attributes
+   */
+  public updateContentMetadata(metadata: Metadata) {
```
bitmovin-kenny: i would suggest `metadataOverrides` as param name

bitmovin-kenny (changes requested):
generally looking good, some nitpicks

src/ts/ContentMetadataBuilder.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    this.contentMetadata.encodedFrameRate = this.metadataOverrides.encodedFrameRate || this.metadata.encodedFrameRate;
+    this.contentMetadata.defaultResource = this.metadataOverrides.defaultResource || this.metadata.defaultResource;
+    this.contentMetadata.streamUrl = this.metadataOverrides.streamUrl || this.metadata.streamUrl;
+
+    return this.contentMetadata;
+  }
+
+  // Those methods should be treated as package private
```
reevn:
```suggestion
  // These methods should be treated as package private
```

src/ts/ConvivaAnalytics.ts
reevn: It's just strange for Austrian brains to see no commas in such a sentence. I think it is valid english syntax though.

Maybe we could do with some commas though?

reevn (approved):
Code looks fine for me. Can't test the functionality live though 👍

src/ts/ConvivaAnalytics.ts
stonko1994: Hm ... actually not. `isValidSession` actually don't check if the metadata are valid, only if a session is active.

Not sure if we need to rename it to solve confusion.

src/ts/ConvivaAnalytics.ts
stonko1994: What about `// Without setting the asset name on the content metadata the SDK will throw errors when we initialize the session.`

bitmovin-kenny (approved):
👍

PR #17, opened 2019-01-30
Add tracking support based on player events in v8

stonko1994:
## Description
Some events changed in player v8 so we need to adapt those in the integration.

Changes which are covered in this PR:
- No stalling events during `play` / `playing`; `seek` / `seeked` and `timeshift` / `timeshifted`
- Workaround for wrong player event in case of a pre-roll ad (The `adbreakstarted` event is triggered before an `play`)
- Added `seek` and `timeshift` tracking again
- Remove outdated flag

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -250,7 +250,6 @@ export class ConvivaAnalytics {
     this.contentMetadata.applicationName = this.config.applicationName || 'Unknown (no config.applicationName set)';
     this.contentMetadata.assetName = this.getAssetName(source);
     this.contentMetadata.viewerId = source.viewerId || this.config.viewerId || null;
-    this.contentMetadata.duration = this.player.getDuration();
     this.contentMetadata.streamType =
       this.player.isLive() ? Conviva.ContentMetadata.StreamType.LIVE : Conviva.ContentMetadata.StreamType.VOD;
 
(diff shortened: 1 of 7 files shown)


src/ts/ConvivaAnalytics.ts
```diff
@@ -416,40 +449,42 @@ export class ConvivaAnalytics {
   };
 
   private onPlay = (event: PlaybackEvent) => {
-    this.debugLog('play', event);
+    this.debugLog('[ Player Event ] play', event);
+
     if (this.isAd) {
       // Do not track play event during ad (e.g. triggered from IMA)
+      console.log('[log] returning cause of ad (onplay)');
```
bitmovin-kenny: is the [log] prefix helpful?
`Ignoring play event because an ad is active` would be easier to understand

src/ts/ConvivaAnalytics.ts
```diff
@@ -366,48 +368,79 @@ export class ConvivaAnalytics {
...
+  private isSessionActive(): boolean {
     return this.sessionKey !== Conviva.Client.NO_SESSION_KEY;
   }
 
-  private onPlaybackStateChanged = (event?: PlayerEventBase) => {
-    if (this.isAd) {
-      // Do not track playback state changes during ad (e.g. triggered from IMA)
+  private onPlaybackStateChanged = (event: PlayerEventBase) => {
+    // Do not track playback state changes during ads, (e.g. triggered from IMA)
+    // or if there is no active session.
```
bitmovin-kenny, line 386: a reason why not to do that would be helpful i guess

bitmovin-kenny (approved):
Looks ok i guess

reevn (approved):
Looks fine to me 👍

src/ts/ConvivaAnalytics.ts
stonko1994: I guess this will be obsolete soon when we add support for `AdInsights` so I leave it for now.

PR #18, opened 2019-02-11
Add possibility to pause resume tracking

stonko1994:
## Description
This PR ads possibility to pause / resume tracking by calling public methods.

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -218,7 +218,7 @@ export class ConvivaAnalytics {
    * @param severity One of FATAL or WARNING
    * @param endSession Boolean flag if session should be closed after reporting the deficiency (Default: true)
    */
-  reportPlaybackDeficiency(message: string, severity: Conviva.Client.ErrorSeverity, endSession: boolean = true) {
+  public reportPlaybackDeficiency(message: string, severity: Conviva.Client.ErrorSeverity, endSession: boolean = true) {
     if (!this.isSessionActive()) {
       return;
     }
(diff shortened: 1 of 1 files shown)


src/ts/ConvivaAnalytics.ts
```diff
@@ -229,6 +229,31 @@ export class ConvivaAnalytics {
...
+  /**
+   * Puts the session state in a notMonitored state.
+   */
+  public pauseTracking(): void {
+    // AdStart is the right way to pause monitoring according to conviva.
+    this.client.adStart(
+      this.sessionKey,
+      Conviva.Client.AdStream.SEPARATE,
+      Conviva.Client.AdPlayer.SEPARATE,
+      Conviva.Client.AdPosition.PREROLL, // Also stops tracking time for VST so PREROLL seems to be sufficient
```
bitmovin-kenny, line 241: what is `VST`?

src/ts/ConvivaAnalytics.ts
```diff
@@ -229,6 +229,31 @@ export class ConvivaAnalytics {
...
+    );
+    this.client.detachPlayer(this.sessionKey);
```
bitmovin-kenny, line 243: shouldn't you update the playerStateManager here?

src/ts/ConvivaAnalytics.ts
stonko1994: `Video Startup Time`

src/ts/ConvivaAnalytics.ts
stonko1994: No because the internal tracking doesn't stop but just don't get propagated to conviva. After the tracking is resumed the same instance of `playerStateManager` will be and should be used.

bitmovin-kenny (approved):
👍

PR #19, opened 2019-02-12
Add tests based on jest

stonko1994:
## Description
This PR will add a basic setup for `jest`.

## Tests
- Content metadata tests (include content metadata overriding)
- Externally session managing tests
- Player events tests (session handling based on player events)


spec/helper/MockHelper.ts
```diff
@@ -0,0 +1,335 @@
...
+
+    return new PlayerMockClass();
+  }
+}
+
+export interface TestingPlayerAPI extends PlayerAPI {
+  eventEmitter: EventEmitter;
+}
+
+interface EventEmitter {
```
unki2aut, line 130: This interface is not really needed, why do we have it?

stonko1994: This make sure that we don't need an any cast when firing events.

unki2aut (approved):
Test setup looks good to me :+1:

jest.config.js
```diff
@@ -0,0 +1,25 @@
...
+
+  // A map from regular expressions to paths to transformers
+  transform: {
+    '^.+\\.tsx?$': 'ts-jest'
+  },
+
+  // A set of global variables that need to be available in all test environments.
+  globals: {
+    window: { },
+    navigator: { }, // TODO: find a way to mock the BrowserUtils
```
protyposis: Just set `navigator.userAgent` to an appropriate value?

spec/helper/PlayerEvent.ts
```diff
@@ -0,0 +1,807 @@
+export enum PlayerEvent {
```
protyposis, line 1: Not a big fan of this. Why can't you use the `PlayerEvent` enum from the player package?

spec/tests/ContentMetadata.spec.ts
```diff
@@ -0,0 +1,331 @@
...
+        describe('it does update', () => {
+          it('encoded frame rate', () => {
+            convivaAnalytics.updateContentMetadata({ encodedFrameRate: 144 });
+
+            expect(clientMock.updateContentMetadata).toHaveBeenLastCalledWith(0, expect.objectContaining({
+              encodedFrameRate: 144,
+            }));
+          });
+
+          it('defaultResrouce', () => {
```
protyposis: Typo

protyposis (changes requested):

spec/helper/PlayerEvent.ts
unki2aut: If we import the enum from `bitmovin-player` we can't mock it off and then the whole player code is executed and all the browser specific variables would have to be mocked as well.

We can't import the enum form the types, as `d.ts` files are not compiled by the `tsc`.

protyposis: Sad but true

jest.config.js
stonko1994: Found a way to mock it properly.

protyposis (approved):

PR #20, opened 2019-02-15
Update readme

stonko1994:
## Description
Update readme with:
- a remark about how consecutive playback with the same player instance is handled
- how to override content metadata

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,31 @@ _See [ConvivaAnalytics.ts](./src/ts/ConvivaAnalytics.ts) for parameter details._
 
 Conviva suggests an timeout of about ~10 seconds and before reporting an error to conviva and providing feedback the user.
 
+#### Content Metadata handling
+
+If you want to override some content metadata attributes you can do so by adding the following:
+
+```js
+let metadataOverrides = {
+  applicationName: 'App Name',
+  custom: {
+    customTag: 'customValue',
+  },
+  encodedFrameRate: 24,
+  // … 
+};
+
+// …
+// Initialize ConvivaAnalytics
+// …
+
+conviva.updateContentMetadata(metadataOverrides);
+```
+
+Those values will be cleaned up after the session is closed.
+
+_See [ConvivaAnalytics.ts](./src/ts/ConvivaAnalytics.ts) for details about more attributes._
+
 #### Consecutive playback
 
 If you want to use the same player instance for multiple playback call `player.unload()` before loading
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,3 +75,13 @@ conviva.reportPlaybackDeficiency('Some Error Message', Conviva.Client.ErrorSever
 _See [ConvivaAnalytics.ts](./src/ts/ConvivaAnalytics.ts) for parameter details._
 
 Conviva suggests an timeout of about ~10 seconds and before reporting an error to conviva and providing feedback the user.
+
+#### Consecutive playback
+
+If you want to use the same player instance for multiple playback call `player.unload()` before loading
```
protyposis: Why can't one just call `load` without `unload`? If this is a necessity in contrast to the plain player, then it should explicitly be addressed here like that, e.g. ``In contrast to the normal** player, you have to call `player.unload()` before loading a new source with `player.load(...)`.``

** find better term (@zachmanc?)

protyposis (changes requested):

README.md
zachmanc: maybe something like...``When using the conviva integration, you must call `player.unload()` before loading a new source with `player.load(...)`. This allows us to properly close the previous session before starting a new one``

Also how do we update metadata across playback sessions? We should provide an example of how to do that

stonko1994: @zachmanc update metadata across playback sessions is currently not supported. Should this be supported?

protyposis: @zachmanc Your phrasing does not explicitly address the API usage difference to the plain player. I think this is important because it means that you cannot simply add the conviva integration while keeping the player loading code unchanged.

protyposis: @stonko1994 I am still wondering why you cannot handle this internally, e.g. close the previous session on a `load` without preceding `unload`.

stonko1994: @protyposis Since we have externally session managing support (creating the session manually and potential even before a source was loaded) I can't just close the session on `sourceLoaded`. So we rely on a clear intention to close the session wich is in the case of multiple playbacks the `sourceUnloaded`.

protyposis: But `PlayerEvent.SourceUnloaded` will be fired in any case, because `load` simply calls `unload` internally.

stonko1994: Also for the initial load call?

EDIT: anyway that will make the readme easier cause the `unload` call is not necessary then.

protyposis: Only if there is a source loaded, so the first `load` will not trigger this event.

stonko1994:
@protyposis @zachmanc I have update the readme. Please have another look.

protyposis (approved):

zachmanc (approved):

PR #22, opened 2019-02-27
Add support for Ad insights

stonko1994:
## Description
This will ad support for ad insights tracking.
To properly track values this feature relies the `AdData` of 8.3.0.

## Tests
- AdTracking
  - Tests about ad tracking for each mode where added or adapted


src/ts/AdExperienceTrackingPlugin.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  private reportBitrate(bitrate: number) {
+    if (!bitrate) {
+      return;
+    }
+
+    let kbpsBitrate = bitrate;
+    // Since the unit isn't defined for adData.bitrate (neither in the Vast Manifest) we assume that values greater
+    // than 10_000 are in bps and convert them to kbps
+    if (bitrate >= 10_000) {
```
reevn, line 101: Guess there is no way to make this future proof? Hope nobody will get the idea to serve 10mbit ads 🙂

src/ts/AdExperienceTrackingPlugin.ts
```diff
@@ -0,0 +1,132 @@
...
+      Conviva.Client.AdTechnology.CLIENT_SIDE;
+
+    // custom takes a javascript object as an argument
+    adMetadata.custom = {
+      // Required
+      'c3.ad.technology': String(adTechnology),
+      'c3.ad.id': ad.id || AdExperienceTrackingPlugin.UNKNOWN_VALUE,
+      'c3.ad.system': adData && adData.adSystem && adData.adSystem.name || AdExperienceTrackingPlugin.UNKNOWN_VALUE,
+      'c3.ad.position': String(this.currentAdBreakPosition),
+      'c3.ad.type': AdExperienceTrackingPlugin.UNKNOWN_VALUE,
```
reevn: What does Conviva need for `type`?

spec/tests/AdTracking.spec.ts
```diff
@@ -0,0 +1,463 @@
+import { AdTrackingMode, ConvivaAnalytics, ConvivaAnalyticsConfiguration } from '../../src/ts/ConvivaAnalytics';
```
offnertho, line 1: why are we using a different test name pattern?

src/ts/ConvivaAnalytics.ts
```diff
@@ -324,6 +349,20 @@ export class ConvivaAnalytics {
 
     // Create a Conviva monitoring session.
     this.sessionKey = this.client.createSession(this.contentMetadataBuilder.build()); // this will make the initial request
+
+    // Init ad tracking with current session key
+    switch (this.config.adTrackingMode) {
+      case AdTrackingMode.AdExperience:
+        this.adTrackingModule = new AdExperienceTrackingPlugin(this.player, this.client, this.sessionKey, this.logger);
```
offnertho: are these `plugins` or `modules`? We should name the variable accordingly.

src/ts/ConvivaAnalytics.ts
```diff
@@ -465,7 +498,11 @@ export class ConvivaAnalytics {
 
     if (playerState) {
       this.debugLog('[ ConvivaAnalytics ] report playback state', playerState);
-      this.playerStateManager.setPlayerState(playerState);
+      if (this.adTrackingModule && this.adTrackingModule.isAdSessionActive()) {
```
offnertho: how can there be no `adTrackingModule`? Shouldn't we check for `isAdActive`

src/ts/ConvivaAnalytics.ts
```diff
@@ -516,7 +548,7 @@ export class ConvivaAnalytics {
   private onVideoQualityChanged = (event: VideoQualityChangedEvent) => {
     // We calculate the bitrate with a divisor of 1000 so the values look nicer
     // Example: 250000 / 1000 => 250 kbps (250000 / 1024 => 244kbps)
```
offnertho: ? What is it now? `/1000` or `/1024`
Also I think the comment is no longer needed, as this info is present in the method name of the `BitrateHelper`

src/ts/ConvivaAnalytics.ts
```diff
@@ -687,25 +700,21 @@ export class ConvivaAnalytics {
     playerEvents.add(this.events.Seeked, this.onSeeked);
     playerEvents.add(this.events.TimeShift, this.onTimeShift);
     playerEvents.add(this.events.TimeShifted, this.onTimeShifted);
+
+    // Ad tracking events
+    playerEvents.add(this.events.AdStarted, this.onAdStarted);
```
offnertho, line 703: are those also removed somewhere?

src/ts/ConvivaAnalytics.ts
```diff
@@ -687,25 +700,21 @@ export class ConvivaAnalytics {
...
-          type: this.events.PlaybackFinished,
-        },
-      );
-      return;
-    }
+  private onAdFinished = (event: AdEvent) => {
+    this.adTrackingModule.adFinished();
+  };
 
+  private onAdBreakStarted = (event: AdBreakEvent) => {
```
offnertho, line 715: can't we initialize the session now on the adbreak started?

src/ts/helper/BitrateHelper.ts
```diff
@@ -0,0 +1,7 @@
+export class BitrateHelper {
```
offnertho, line 1: unit test?

src/ts/helper/ObjectUtils.ts
```diff
@@ -0,0 +1,23 @@
+import { EventAttributes } from '../ConvivaAnalytics';
```
offnertho, line 1: unit test?

offnertho (changes requested):
Seems legit in general. 

Some small remarks.

src/ts/AdExperienceTrackingPlugin.ts
stonko1994: I just wanted to read that up and it seems like they changed something here.

The `ad.type` is gone but now there is an `c3.ad.isSlate`. 
```
A boolean value that indicates whether this ad is a Slate or not. Set to "true" for Slate and "false" for a regular ad. By default, set to "false".
```

spec/tests/AdTracking.spec.ts
stonko1994: Different to which one?

src/ts/ConvivaAnalytics.ts
stonko1994: I would go for `this.adTrackingPlugin`

src/ts/ConvivaAnalytics.ts
stonko1994: This is to be bullet proof. In general your statement is correct.

But we bind the player events in the constructor but `this.adTrackingModule` will only be set when the session get initialized. So it could happen that this method is called before we have an active session.

-> No we can't initialize the `this.adTrackingModule` in the constructor because it needs an active tracking session.

src/ts/ConvivaAnalytics.ts
stonko1994: it's `/ 1000` but you are right its obsolete and also duplicated in the `BitrateHelper`

src/ts/ConvivaAnalytics.ts
stonko1994: Yes, in `unregisterPlayerEvents`

src/ts/ConvivaAnalytics.ts
stonko1994: Hm what do you mean actually?

This is still needed as the event order is incorrect as the comment is pointing out.

spec/tests/AdTracking.spec.ts
offnertho: `AdTrackingSpec.ts`, the one we use in the player. But I guess this is fine.

src/ts/AdExperienceTrackingPlugin.ts
```diff
@@ -0,0 +1,132 @@
...
+      'c3.ad.adStitcher': isYospaceUsed ? 'Yospace' : AdExperienceTrackingPlugin.UNKNOWN_VALUE,
+    };
+
+    const addValueIfPresent = (key: string, value: string) => {
+      if (value) {
+        adMetadata.custom[key] = value;
+      }
+    };
+
+    addValueIfPresent('c3.ad.creativeId', adData && adData.creative && adData.creative.adId);
```
offnertho: can u please move all the part which is relying on having `adData` into a separate function.

src/ts/ConvivaAnalytics.ts
offnertho: Nevermind.

spec/tests/AdTracking.spec.ts
stonko1994: Ah I thought you mean within the project. Anyway I just used the `jest` default pattern.

reevn (approved):

offnertho (approved):

PR #23, opened 2019-03-04
Fix missing bitrate tracking on session initialization

stonko1994:
## Description
Since the first `videoPlaybackQualityChanged` event gets fired before the playback ever started the event can't be tracked 'cause no conviva tracking session is open at that point.

If the `videoPlaybackQuality` never changes again we never report a bitrate. To fix this issue (and improve) the bitrate will be remembered and set on session initialization.

## Tests
- Tests about bitrate tracking are added

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -62,6 +62,10 @@ export class ConvivaAnalytics {
   // Attributes needed to workaround wrong event order in case of a pre-roll ad. (See #onAdBreakStarted for more info)
   private adBreakStartedToFire: AdBreakEvent;
 
+  // Needed to workaround wrong event order in case of a video-playback-quality-change event. (See
+  // #onVideoQualityChanged for more info)
+  private lastSeenBitrate: number;
+
   // Since there are no stall events during play / playing; seek / seeked; timeShift / timeShifted we need
   // to track stalling state between those events. To prevent tracking eg. when seeking in buffer we delay it.
   private stallTrackingTimout: Timeout = new Timeout(ConvivaAnalytics.STALL_TRACKING_DELAY_MS, () => {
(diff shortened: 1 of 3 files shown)


src/ts/ConvivaAnalytics.ts
```diff
@@ -62,6 +62,10 @@ export class ConvivaAnalytics {
   // Attributes needed to workaround wrong event order in case of a pre-roll ad. (See #onAdBreakStarted for more info)
   private adBreakStartedToFire: AdBreakEvent;
 
+  // Needed to workaround wrong event order in case of a video-playback-quality-change event. (See
+  // #onVideoQualityChanged for more info)
```
protyposis: The `#` notation does not exist in JS/TS. It's always a `.`, or in this case, `{@link onVideoQualityChanged}`.

src/ts/ConvivaAnalytics.ts
```diff
@@ -509,14 +519,19 @@ export class ConvivaAnalytics {
...
-    }
     // We calculate the bitrate with a divisor of 1000 so the values look nicer
     // Example: 250000 / 1000 => 250 kbps (250000 / 1024 => 244kbps)
     const bitrateKbps = Math.round(event.targetQuality.bitrate / 1000);
-    console.warn('go video quality changed ', this.sessionKey, bitrateKbps);
 
+    if (!this.isSessionActive()) {
+      // Since the first videoPlaybackQualityChanged event comes before playback ever started we need to store the
+      // value and use it for tracking when initializing the session.
+      // TODO: remove this workaround when the player event order is fixed
```
protyposis, line 534: The player event order probably won't be fixed. It selects a bitrate before playback starts, and it can't do it the other way around - because with which bitrate would it start then?

protyposis (approved):

src/ts/ConvivaAnalytics.ts
stonko1994: Hm but it feels weird that a `*Playback*Changed` event is fired without active `playback`.

offnertho (approved):

PR #24, opened 2019-03-04
Fix issue that reported values for seek events need to be integer values

stonko1994:
## Description
Values reported to conviva via seek tracking api need to be integer values.

## Fix
Round float values to integers.

## Tests
Seek tests adapted.


bitmovin-kenny (approved):
Makes little sense to me to round the seek targets, but if it's required 👍

PR #26, opened 2019-03-07
Prepare npm publishing

stonko1994:
## Description
This PR will add required changes to enable npm publishing.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -1,3 +1,4 @@
+const CreateFileWebpack = require('create-file-webpack');
 const npmPackage = require('./package.json');
 const path = require('path');
 
@@ -25,7 +26,7 @@ module.exports = {
     extensions: ['.tsx', '.ts', '.js']
   },
   output: {
-    filename: 'bitmovinplayer-analytics-conviva.js',
+    filename: 'bitmovin-player-analytics-conviva.js',
     umdNamedDefine: true,
     path: path.resolve(__dirname, 'dist'),
     library: {
@@ -36,4 +37,14 @@ module.exports = {
     libraryTarget: 'umd'
   },
   devtool: 'source-map',
+  plugins: [
+    new CreateFileWebpack({
+      // path to folder in which the file will be created
+      path: './dist',
+      // file name
+      fileName: 'bitmovin-player-analytics-conviva.d.ts',
+      // content of the file
+      content: 'export * from \'./lib/index\''
+    }),
+  ],
 };
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -44,7 +44,7 @@ module.exports = {
       // file name
       fileName: 'bitmovin-player-analytics-conviva.d.ts',
       // content of the file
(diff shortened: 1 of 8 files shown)


protyposis (commented):
There are a few odd things with the `package.json` here:
 * This cannot be used as local package because the `main` and `types` entries point to files that do not exist, i.e. they only exist after the file has been copied to `dist`
 * The other way round, the `scripts` entries do not make sense within the `dist` folder and npm package
 * Why does `main` not point to `bitmovinplayer-analytics-conviva.js`? Why is there no `bitmovinplayer-analytics-conviva.d.ts`?

stonko1994:
Re: 
> This cannot be used as local package because the main and types entries point to files that do not exist, i.e. they only exist after the file has been copied to dist

Where should they point then? As I understood it they should point to the `dist` (after transpiling) files. This is also the same case in the UI package

---

Re:
> The other way round, the scripts entries do not make sense within the dist folder and npm package

True

---

Re: 
> Why does main not point to bitmovinplayer-analytics-conviva.js? Why is there no bitmovinplayer-analytics-conviva.d.ts?

Why do we need the `bitmovinplayer-analytics-conviva.d.ts`? There are all types in the files included in `d.ts` files in the `dist` folder.

> Why does main not point to bitmovinplayer-analytics-conviva.js

Should it point to that? I'm not sure if there is a difference? (also taken from the UI)

protyposis:
> Where should they point then? As I understood it they should point to the dist (after transpiling) files. This is also the same case in the UI package

They should point to files that exist after running the build task.

> Why do we need the `bitmovinplayer-analytics-conviva.d.ts`? There are all types in the files included in `d.ts` files in the `dist` folder.

Because TS does automatically load types from `d.ts` files that live along `js` files. The file can simply reexport from `index.d.ts`.

> Should it point to that? I'm not sure if there is a difference? (also taken from the UI)

Yes the compiled production build should be the default. Everybody is still free to import single files explicitly from `lib` if there is any reason for that. (The UI isn't perfect either ;) )

protyposis (changes requested):
As outlined above.

reevn (commented):
I'm just a silent watcher because this stuff interests me and I want to learn more about it. Don't wait for any real review from me 🙂

webpack.common.js
```diff
@@ -36,4 +37,14 @@ module.exports = {
...
   },
   devtool: 'source-map',
+  plugins: [
+    new CreateFileWebpack({
+      // path to folder in which the file will be created
+      path: './dist',
+      // file name
+      fileName: 'bitmovin-player-analytics-conviva.d.ts',
+      // content of the file
+      content: 'export * from \'./lib/index\''
```
protyposis: Trailing `;` missing

protyposis (approved):
Looking good now! (just a minor cosmetic issue)

unki2aut (approved):
LGTM :+1:

PR #27, opened 2019-03-08
Add npmignore file to remove files from package

stonko1994:
## Description
This will remove some unnecessary files from the package.

Also it will ensure that the `dist` folder is included when publishing. It would be ignored because it is ignored in the `.gitignore` file which is used per default.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,3 @@
-!dist
 .idea
 spec
 jest.config.js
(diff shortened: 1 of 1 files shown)


.npmignore
```diff
@@ -0,0 +1,7 @@
+!dist
```
dweinber: That's not necessary, everything which is not part of the rest of the list is not ignored anyway

dweinber (approved):

PR #28, opened 2019-03-27
Chromecast support

stonko1994:
## Description
We currently don't have dedicated chromecast support within the integration. It was able to handle that with the methods we provided for external session managing.

This PR adds support for chromecast.

## Tests
New tests about testing casting behaviour were added.


reevn (approved):
Code LGTM 👍

PR #29, opened 2019-04-24
Example on how to use the integration on chrome cast

stonko1994:
## Description
Currently we have no example on how to use the integration on a chrome cast device

## Fix
- This PR adds an example of how to use the integration within a custom cast receiver app.
- Stop the tracking session from the sender app on `CastWaitingForDevice` as this is the event when the user selected a device.

## Tests
Adopted to the new event.

## Manual Tests
- Create a custom receiver app and use the new `receiverApp.html` file
- Ensure that the session of the sender app gets closed and a new one gets created from the receiver app.

### DIFF
--- a/example/.gitignore
+++ b/example/.gitignore
@@ -1,6 +1,1 @@
-*
-!.gitignore
-!README.md
-!index.html
-!script.js
-!style.css
+conviva-core-sdk.min.js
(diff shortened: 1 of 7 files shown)


jmsn (approved):
Tested and LGTM 👍

PR #30, opened 2019-04-29
Content metadata support on chromecast

stonko1994:
## Description
Currently you have to update content metadata on the sender app as well as on the receiver app event the same values should be used.

## Solution
Introduce a proper way on how to send data from the sender app to the receiver app. This lead to the opportunity that the content metadata attributes just need to be set once and also can be set directly via the sender app.

## Changes
- Close the cast session on the sender app immediately when the user clicks on a cast device instead of the `CastStarted` event (Unreleased)

## Tests
New tests added

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,32 @@ The recommended version of the Conviva SDK is 2.146.0.36444.** See [CHANGELOG](C
  
 ### Advanced Usage
 
+#### Chromecast support
+
+_If you are not familiar with how to add chromecast support we suggest to look at our [public sample](https://github.com/bitmovin/bitmovin-player-web-samples/tree/master/castReceiver) first._
+
+If you would like to add the conviva integration to your chromecast you can have a look at our [custom receiver app example](./example/chromecast/receiverApp.html) with conviva integration.
+
+As the player itself does not notify listeners on the receiver app itself if the session will be stopped you need to add a listener to the `onShutdown` and end the session manually.
+
+For doing so you need to add following lines:
+
+```js
+cast.receiver.CastReceiverManager.getInstance().onShutdown = () => {
+  convivaAnalytics.endSession();
+}
+```
+
+If you are using the content metadata overriding feature you will have to add following line within the `CastMetadataListener` callback:
+
+```js
+convivaAnalytics.handleCastMetadataEvent(metadata);
+```
+
+This will ensure that all your content metadata attributes are also present in the session of the receiver app.
+It will also enable that you can use `updateContentMetadata` on the sender and it will be propagated to the receiver app.
+
+
 #### VPF tracking
 
 If you would like to track custom VPF (Video Playback Failures) events when no actual player error happens (e.g. 
--- a/README.md
+++ b/README.md
@@ -57,9 +57,9 @@ _If you are not familiar with how to add chromecast support we suggest to look a
 
 If you would like to add the conviva integration to your chromecast you can have a look at our [custom receiver app example](./example/chromecast/receiverApp.html) with conviva integration.
 
-As the player itself does not notify listeners on the receiver app itself if the session will be stopped you need to add a listener to the `onShutdown` and end the session manually.
+As the player does not notify listeners on the receiver app itself if the session will be stopped you need to add a listener to the `onShutdown` and end the session manually.
 
-For doing so you need to add following lines:
+In order to do so you need to add following lines:
 
 ```js
 cast.receiver.CastReceiverManager.getInstance().onShutdown = () => {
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -51,6 +51,32 @@ The recommended version of the Conviva SDK is 2.146.0.36444.** See [CHANGELOG](C
...
+
+For doing so you need to add following lines:
+
+```js
+cast.receiver.CastReceiverManager.getInstance().onShutdown = () => {
+  convivaAnalytics.endSession();
+}
+```
+
+If you are using the content metadata overriding feature you will have to add following line within the `CastMetadataListener` callback:
```
reevn:
```suggestion
If you are using the content metadata overriding feature you will have to add the following line within the `CastMetadataListener` callback:
```

README.md
```diff
@@ -51,6 +51,32 @@ The recommended version of the Conviva SDK is 2.146.0.36444.** See [CHANGELOG](C
...
+
+```js
+convivaAnalytics.handleCastMetadataEvent(metadata);
+```
+
+This will ensure that all your content metadata attributes are also present in the session of the receiver app.
+It will also enable that you can use `updateContentMetadata` on the sender and it will be propagated to the receiver app.
```
reevn:
```suggestion
It will also enable you to use `updateContentMetadata` on the sender and it will be propagated to the receiver app.
```

README.md
```diff
@@ -51,6 +51,32 @@ The recommended version of the Conviva SDK is 2.146.0.36444.** See [CHANGELOG](C
...
 
+#### Chromecast support
+
+_If you are not familiar with how to add chromecast support we suggest to look at our [public sample](https://github.com/bitmovin/bitmovin-player-web-samples/tree/master/castReceiver) first._
+
+If you would like to add the conviva integration to your chromecast you can have a look at our [custom receiver app example](./example/chromecast/receiverApp.html) with conviva integration.
+
+As the player itself does not notify listeners on the receiver app itself if the session will be stopped you need to add a listener to the `onShutdown` and end the session manually.
+
+For doing so you need to add following lines:
```
reevn:
```suggestion
In order to do so you need to add following lines:
```

README.md
```diff
@@ -51,6 +51,32 @@ The recommended version of the Conviva SDK is 2.146.0.36444.** See [CHANGELOG](C
  
 ### Advanced Usage
 
+#### Chromecast support
+
+_If you are not familiar with how to add chromecast support we suggest to look at our [public sample](https://github.com/bitmovin/bitmovin-player-web-samples/tree/master/castReceiver) first._
+
+If you would like to add the conviva integration to your chromecast you can have a look at our [custom receiver app example](./example/chromecast/receiverApp.html) with conviva integration.
+
+As the player itself does not notify listeners on the receiver app itself if the session will be stopped you need to add a listener to the `onShutdown` and end the session manually.
```
reevn:
```suggestion
As the player does not notify listeners on the receiver app itself if the session will be stopped you need to add a listener to the `onShutdown` and end the session manually.
```

src/ts/ConvivaAnalytics.ts
```diff
@@ -423,6 +446,20 @@ export class ConvivaAnalytics {
     this.client.updateContentMetadata(this.sessionKey, this.contentMetadataBuilder.build());
   }
 
+  // This will propagate content metadata overrides to the conviva instance of the receiver app
+  private propagateOverridesToReceiver(metadataOverrides?: Metadata) {
+    metadataOverrides = metadataOverrides ? metadataOverrides : this.contentMetadataBuilder.getOverrides();
```
reevn: I'd prefer a classic logical OR.
```suggestion
    metadataOverrides = metadataOverrides || this.contentMetadataBuilder.getOverrides();
```

src/ts/ConvivaAnalytics.ts
```diff
@@ -447,11 +484,20 @@ export class ConvivaAnalytics {
     this.client.releasePlayerStateManager(this.playerStateManager);
 
     this.sessionKey = Conviva.Client.NO_SESSION_KEY;
-    this.contentMetadataBuilder.reset();
-
     this.lastSeenBitrate = null;
+
+    // As the session could be continued after casting we can't reset the contentMetadataBuilder here as we would loose
```
reevn:
```suggestion
    // As the session could be continued after casting we can't reset the contentMetadataBuilder here as we would lose
```

src/ts/ConvivaAnalytics.ts
```diff
@@ -319,6 +327,21 @@ export class ConvivaAnalytics {
     }
   }
 
+  private internalSetOverrides(metadataOverrides: Metadata) {
```
reevn: Why the `internal` prefix?

reevn (changes requested):

src/ts/ConvivaAnalytics.ts
stonko1994: Actually just to match the naming scheme with `internalEndSession` / `internalInitializeSession` but it should then be `internalUpdateContentMetadata`.

I will rename it to `internalUpdateContentMetadata` to be consistent.

reevn (approved):
LGTM 👍

jmsn (approved):

PR #34, opened 2019-07-03
Do not initialize session internally when endSession is called externally

jonathanhperry:
…ally

If a developer chooses to end the Conviva session explicitly via the `endSession()` call, we should not re-initialize the session when certain player events fire and instead should only allow the external caller to re-initialize the session via `initializeSession` at their discretion.

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -186,7 +186,9 @@ export class ConvivaAnalytics {
    * Ends the current conviva tracking session.
    * Results in a no-opt if there is no active session.
    *
-   * Warning: The integration can only be validated without external session managing. So when using this method we can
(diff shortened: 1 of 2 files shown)


src/ts/ConvivaAnalytics.ts
```diff
@@ -189,6 +196,7 @@ export class ConvivaAnalytics {
```
stonko1994, line 198: _I can't comment on the documentation of this method._

Please add a remark to the method documentation that no session will be created automatically anymore after calling this.

stonko1994 (commented):
Nice 👍

src/ts/ConvivaAnalytics.ts
jonathanhperry: Thanks, added a note to the `endSession` method documentation

stonko1994 (approved):

PR #35, opened 2019-07-09
Change updateContentMetadata loglevel

jonathanhperry:
The "WARNING" level of the message when the session is not active can throw off developers as it implies something may not be setup correctly.

Since we are still updating the content metadata but wait until session initialization, changing the log level to "DEBUG" seems more appropriate.


stonko1994 (commented):
Changes makes sense 👍

please also add a changelog even the change is that small. (Please also sneek in a changelog entry for #34, forgot it there)

jonathanhperry:
changelog updated to include both changes

stonko1994 (approved):

PR #45, opened 2020-01-28
Add a ConvivaAnalyticsConfiguration option to set DeviceCategory

mckennajones:
After going back and forth with Conviva we determined that the best approach to handling Device Category would be to allow an option for setting the category. This value will then be used within the Conviva SDK to influence their parsing of the user agent string.

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -21,6 +21,7 @@ import { BasicAdTrackingPlugin } from './BasicAdTrackingPlugin';
 import { BitrateHelper } from './helper/BitrateHelper';
 
 import { ArrayUtils } from 'bitmovin-player-ui/dist/js/framework/arrayutils';
+import { DeviceMetadata } from './DeviceMetadata';
 
 type Player = PlayerAPI;
 
@@ -139,12 +140,16 @@ export class ConvivaAnalytics {
     this.logger = new Html5Logging();
     this.sessionKey = Conviva.Client.NO_SESSION_KEY;
 
+    const deviceMetadata: DeviceMetadata = {
+      deviceCategory: this.config.deviceCategory,
+    };
+
     const systemInterface = new Conviva.SystemInterface(
       new Html5Time(),
       new Html5Timer(),
       new Html5Http(),
       new Html5Storage(),
-      new Html5Metadata(this.config.deviceCategory),
+      new Html5Metadata(deviceMetadata),
       this.logger,
     );
 
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -21,7 +21,7 @@ import { BasicAdTrackingPlugin } from './BasicAdTrackingPlugin';
 import { BitrateHelper } from './helper/BitrateHelper';
 
 import { ArrayUtils } from 'bitmovin-player-ui/dist/js/framework/arrayutils';
-import { DeviceMetadata } from './DeviceMetadata';
+import { DeviceMetadata } from './Html5Metadata';
 
 type Player = PlayerAPI;
 
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -134,6 +134,7 @@ export class ConvivaAnalytics {
 
     // Set default config values
     this.config.debugLoggingEnabled = this.config.debugLoggingEnabled || false;
+    this.config.deviceCategory = this.config.deviceCategory || Conviva.Client.DeviceCategory.WEB;
 
     this.logger = new Html5Logging();
     this.sessionKey = Conviva.Client.NO_SESSION_KEY;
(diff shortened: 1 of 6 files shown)


src/ts/Html5Metadata.ts
```diff
@@ -1,6 +1,11 @@
 import Client = Conviva.Client;
 
 export class Html5Metadata implements Conviva.MetadataInterface {
+  deviceCategory?: Client.DeviceCategory;
+
+  constructor(deviceCategory?: Client.DeviceCategory) {
```
jonathanhperry: Is there a particular reason to have the DeviceCategory as a constructor argument instead of just as a settable property in this class?

src/ts/ConvivaAnalytics.ts
```diff
@@ -128,6 +134,7 @@ export class ConvivaAnalytics {
 
     // Set default config values
     this.config.debugLoggingEnabled = this.config.debugLoggingEnabled || false;
+    this.config.deviceCategory = this.config.deviceCategory || Conviva.Client.DeviceCategory.WEB;
```
jonathanhperry, line 138: I'd probably prefer to set a default DeviceCategory in `Html5Metadata` class. I feel like if ConvivaAnalytics is refactored and someone removes this OR condition, then DeviceCategory would no longer be sent over in the Conviva beacon.

I'd consider seeing if `this.config.deviceCategory` is set, and if so, to set the property on `Html5Metadata` otherwise defer to the class default.

jonathanhperry (commented):
Might also be worth adding a test case to ensure DeviceCategory gets set correctly

src/ts/Html5Metadata.ts
mckennajones: Not necessarily, only to make instantiating the class a bit easier. Can you think of a reason to not have it in the constructor?

src/ts/ConvivaAnalytics.ts
mckennajones: Makes sense, I updated that.

src/ts/Html5Metadata.ts
jonathanhperry: Opens up to a future where the constructor starts to be overloaded whenever additional properties can be modified. Constructor changes can be breaking changes or can result in lots of overloaded constructors to avoid breaking changes.

Generally prefer to use a setter if a property can be _optionally_ set vs. constructor assignments

src/ts/Html5Metadata.ts
```diff
@@ -59,7 +64,7 @@ export class Html5Metadata implements Conviva.MetadataInterface {
 
   // Relying on HTTP user agent string parsing on the Conviva Platform.
   getDeviceCategory(): Conviva.Client.DeviceCategory {
-    return null;
+    return this.deviceCategory || Conviva.Client.DeviceCategory.WEB;
```
stonko1994: Let's move the default value into `ConvivaAnalytics`. We have a `// Set default config values` in the constructor of `ConvivaAnalytics`. We can set the config option to `WEB` there.

src/ts/Html5Metadata.ts
stonko1994: Or we could pass it as a config object via the constructor.

Creating an interface `DeviceMetadata` or maybe `ClientMetadata` which has a property `deviceCategory` which is mandatory and pass it via the constructor as a `metadata` property.

The problem I see with the getter is that we are implementing an interface `Conviva.MetadataInterface` which does not define those getters. So we can't implement against that interface when using an instance of `Html5Metadata `

stonko1994 (changes requested):
- Please add a changelog entry
- Please close the old PR

mckennajones:
@stonko1994 let me know if my use of the new `DeviceMetadata` interface aligns with what you had in mind.

jonathanhperry (approved):

src/ts/Html5Metadata.ts
```diff
@@ -1,10 +1,11 @@
 import Client = Conviva.Client;
+import { DeviceMetadata } from './DeviceMetadata';
 
 export class Html5Metadata implements Conviva.MetadataInterface {
-  deviceCategory?: Client.DeviceCategory;
+  metadata?: DeviceMetadata;
 
-  constructor(deviceCategory?: Client.DeviceCategory) {
-    this.deviceCategory = deviceCategory;
+  constructor(metadata?: DeviceMetadata) {
```
stonko1994: Make this non-optional now

src/ts/DeviceMetadata.ts
```diff
@@ -0,0 +1,4 @@
+
+export interface DeviceMetadata {
```
stonko1994: I would move this to the `Html5Metadata.ts` file as those belong together. Just add this `export` on top (after the imports)

src/ts/Html5Metadata.ts
```diff
@@ -1,10 +1,11 @@
 import Client = Conviva.Client;
+import { DeviceMetadata } from './DeviceMetadata';
 
 export class Html5Metadata implements Conviva.MetadataInterface {
-  deviceCategory?: Client.DeviceCategory;
+  metadata?: DeviceMetadata;
```
stonko1994: also non-optional now.

stonko1994 (changes requested):

zachmanc (approved):

stonko1994 (approved):

PR #46, opened 2020-02-04
Release/3.1.0

mckennajones:
@stonko1994, are all of these ad tracking changes safe to be released? It looks like develop got quite a bit ahead of master.


stonko1994:
@mckennajones No unfortunately not. The whole `adInsights` never got into validation

PR #49, opened 2020-02-27
Extract ad insights from dev

stonko1994:
### Description
This PR will remove all `AdInsights` related code as we never went into validation with it. The `AdInsights` code still exists on a separate [branch](https://github.com/bitmovin/bitmovin-player-analytics-conviva/tree/feature/ad-insights)


package.json
```diff
@@ -24,7 +24,7 @@
   "license": "ISC",
   "devDependencies": {
     "@types/jest": "^24.0.1",
-    "bitmovin-player": "^8.6.0",
+    "bitmovin-player": "8.6.0",
```
zachmanc, line 27: Doesn't it make more sense to keep this as `^`

zachmanc (approved):

package.json
stonko1994: I thought about it too but we had it in the past that types were changing and therefore we needed to adapt code. Especially in the unit tests.

PR #50, opened 2020-02-28
Revert chromecast support

stonko1994:
### Description

This PR will remove all `chromecast` support related code as we never went into validation with it. The code still exists on a separate [branch](https://github.com/bitmovin/bitmovin-player-analytics-conviva/tree/feature/chromecast-support)


mckennajones (approved):
✅

zachmanc:
@jonathanhperry Do you think this would impact iStreamPlanet?

jonathanhperry:
iSP are still on 3.0.2 of this plug-in which is quite old but includes a hotfix to the double session issue they found.

I don't think reverting this affects them today in production.

PR #51, opened 2020-03-02
Report playing state after ad break is finished

mckennajones:
Small fix to send the playing state after we receive the `adBreakFinished` event﻿


stonko1994 (commented):
- Please add a chanagelog entry
- Did you try what happens after a post-roll ad 🤔 I think it should be fine but just to be sure.

mckennajones:
@stonko1994 Changelog added. For the post-roll, [we check for it here](https://github.com/bitmovin/bitmovin-player-analytics-conviva/blob/master/src/ts/ConvivaAnalytics.ts#L725), and end the session. So it still behaves as expected.

stonko1994 (approved):

PR #66, opened 2020-12-22
Conviva JS SDK Upgrade to 4.x (4.0.15)

turbidwater:


dweinber:
@turbidwater @bknittle The CI is failing, looks like several unit tests are failing now. Could you please check this?

dweinber:
@gkolb could you please sign the [Contributor License Agreement](https://cla-assistant.io/bitmovin/bitmovin-player-analytics-conviva?pullRequest=66)? Thanks!

turbidwater:
Hi @dweinber The unit test issue should now be resolved and Greg has now signed the CLA with both accounts.

src/ts/ConvivaAnalytics.ts
```diff
@@ -79,7 +76,7 @@ export class ConvivaAnalytics {
   // Since there are no stall events during play / playing; seek / seeked; timeShift / timeShifted we need
   // to track stalling state between those events. To prevent tracking eg. when seeking in buffer we delay it.
   private stallTrackingTimout: Timeout = new Timeout(ConvivaAnalytics.STALL_TRACKING_DELAY_MS, () => {
-    this.playerStateManager.setPlayerState(Conviva.PlayerStateManager.PlayerState.BUFFERING);
+    // this.playerStateManager.setPlayerState(Conviva.PlayerStateManager.PlayerState.BUFFERING);
```
dweinber: Why is this commented out? If we don't need it anymore, can we delete this whole code block? Why don't we need this anymore?

src/ts/ConvivaAnalytics.ts
```diff
@@ -110,35 +112,38 @@ export class ConvivaAnalytics {
...
+    callbackFunctions[Conviva.Constants.CallbackFunctions.GET_EPOCH_TIME_IN_MS] = new Html5Time().getEpochTimeMs;
 
-    const clientSettings = new Conviva.ClientSettings(customerKey);
 
+    // If gatewayUrl exists init in production mode for session monitoring. Otherwise init for production
     if (config.gatewayUrl) {
-      clientSettings.gatewayUrl = config.gatewayUrl;
+      const settings = {} as {[key: string]: string| number};
+      settings[Conviva.Constants.GATEWAY_URL] = config.gatewayUrl;
+      settings[Conviva.Constants.LOG_LEVEL] = Conviva.Constants.LogLevel.DEBUG;
```
dweinber: Shouldn't we use `ConvivaAnalyticsConfiguration.debugLoggingEnabled` no matter to set this? And no matter if a `config.gatewayUrl` was provided?

src/ts/ConvivaAnalytics.ts
```diff
@@ -196,7 +202,7 @@ export class ConvivaAnalytics {
    * @param eventAttributes a string-to-string dictionary object with arbitrary attribute keys and values
    */
   public sendCustomApplicationEvent(eventName: string, eventAttributes: EventAttributes = {}): void {
-    this.client.sendCustomEvent(Conviva.Client.NO_SESSION_KEY, eventName, eventAttributes);
+    // this.client.sendCustomEvent(Conviva.Client.NO_SESSION_KEY, eventName, eventAttributes);
```
dweinber: Why is this commented out? If we don't need it anymore, can we delete this whole code block? Why don't we need this anymore? Is this due to the comment on L222 (`// TODO: Look further into custom events. The API from 2.x -> 4.x appears to have lost the option for custom events`)?

src/ts/ConvivaAnalytics.ts
```diff
@@ -461,12 +466,14 @@ export class ConvivaAnalytics {
     switch (event.type) {
       case this.events.Play:
       case this.events.Seek:
+        this.convivaVideoAnalytics.reportPlaybackMetric(Conviva.Constants.Playback.SEEK_STARTED);
```
dweinber: Why do we handle `seek` and `timeshift` differently? Conceptually, those are the same (just VOD vs Live content).
And shouldn't we call `this.stallTrackingTimout.start()` here as well? We're also reseting that timeout (`is.stallTrackingTimout.clear()`) in the `Playing` event, even though we don't start this timer. Which then leads to the follow up question if it's not stopped on purpose in the `Seeked` event. 

I assume this was either overlooked or there's a good reason behind those changes, but might be worth some code comments around Play/Playing, Seek/Seeked, TimeShift/TimeShifted?

src/ts/ConvivaAnalytics.ts
```diff
@@ -346,15 +349,18 @@ export class ConvivaAnalytics {
...
-    this.playerStateManager.setPlayerVersion(this.player.version);
-
     this.buildContentMetadata();
 
     // Create a Conviva monitoring session.
-    this.sessionKey = this.client.createSession(this.contentMetadataBuilder.build()); // this will make the initial request
+    this.convivaVideoAnalytics = Conviva.Analytics.buildVideoAnalytics();
+    this.convivaVideoAnalytics.reportPlaybackRequested(this.contentMetadataBuilder.build());
+    this.sessionKey = this.convivaVideoAnalytics.getSessionId();
+    this.convivaVideoAnalytics.setCallback(() => {
```
dweinber, line 359: When (and if more than once, how often) is this callback called? Is this some kind of heartbeat?

spec/helper/MockHelper.ts
```diff
@@ -19,11 +19,121 @@ export namespace MockHelper {
     global.Conviva.SystemFactory = jest.fn().mockImplementation();
     global.Conviva.ClientSettings = jest.fn().mockImplementation();
     global.Conviva.ContentMetadata = jest.fn().mockImplementation();
+
     global.Conviva.ContentMetadata.StreamType = {
       LIVE: 'live',
       VOD: 'vod',
       UNKNOWN: 'unknown',
     };
+    global.Conviva.Constants = {
```
dweinber, line 28: Would be great to have a consistent code style, here we have strings with `'`, `"` and `` ` `` mixed (the last one could sometimes make sense, but the occurrence I found had just static string data). 

Out of scope for this PR, but maybe we should consider adding `prettier` or similar tooling to take care of those things in the future?

spec/tests/PlayerEvents.spec.ts
```diff
@@ -62,59 +42,44 @@ describe('player event tests', () => {
...
-            done();
-          }, 120);
-        });
         describe('durring playback', () => {
           beforeEach(() => {
             playerMock.eventEmitter.firePlayEvent();
             playerMock.eventEmitter.firePlayingEvent();
           });
 
-          it('delayed after seek', () => {
```
dweinber, line 103: Nice catch, test cases without `expect`ations don't make any sense at all anyway.

dweinber (changes requested):
Looks mostly good, but there are a few points I'd like to be clarified before merging this.

Would be great to have a (brief) PR description in the future so everyone immediately knows the context when looking at this.

src/ts/ConvivaAnalytics.ts
gregdolby: This was unintentional. I'll update this

src/ts/ConvivaAnalytics.ts
gregdolby: I can remove this and leave just the development init call with GATEWAY_URL and LOG_LEVEL

src/ts/ConvivaAnalytics.ts
gregdolby: I'll remove the code block. I am not sure the proper way to send custom events. I posted on the conviva forum asking for a public API because their tutorial is lacking. The public documentation they have doesn't mention anything about custom events

src/ts/ConvivaAnalytics.ts
gregdolby: This was accidental I will remove these and verify seek events are firing correctly from player events.

src/ts/ConvivaAnalytics.ts
gregdolby: According to Conviva docs this fires every one 1 and this is the way they recommend implementing it.

From Conviva Docs:
'There are certain metrics which do not have events, hence Conviva provides a callback to update these values at a certain frequency. Once player has been initialized, implement an callback function for the player so that the VideoAnalytics can get continuous measures of the player such as, play head time, buffer length, and rendered framerate, and other measurements.'

gregdolby:
@dweinber I went through all the comments and made fixes and explained reasonings where necessary. Let me know if you have any questions

src/ts/ConvivaAnalytics.ts
dweinber: I think you didn't get my point here, sorry if I wasn't clear enough. My point is: We should leverage `this.config.debugLoggingEnabled` to decide if debug logging should be enabled or not. The value is set based on the customer's config or if no value is provided turns the logs off, see L114 in this file (https://github.com/bitmovin/bitmovin-player-analytics-conviva/blob/43d4bc06c1da18a14b4b39602569161797c00dc7/src/ts/ConvivaAnalytics.ts#L114).

src/ts/ConvivaAnalytics.ts
dweinber: Makes sense, thanks for the feedback.

src/ts/ConvivaAnalytics.ts
dweinber: I see, thanks 👍

dweinber:
Thanks, @gregrealeyes. One point is still open, I think we had a small misunderstanding there (see my new comment in the thread). I marked the remaining points as resolved.

Unfortunately, the last changes made the build fail again. Please always make sure the build always succeeds before requesting (re-)reviews.

dweinber (changes requested):
Hey @turbidwater,
Happy New Year!

The latest code changes look good, thanks! Unfortunately, there's an error that breaks the test execution, thus the build. If you get this fixed, I think we're good to go.

dweinber (approved):
Not a fan of deleting test cases in general but I assume there are good reasons 😉 
Looks good to me!

dweinber:
@turbidwater could you please update the PR title and description to be more meaningful? Thanks!

dweinber (changes requested):
Missed this previously: Please add a Changelog entry for the changes you did.

src/ts/ConvivaAnalytics.ts
```diff
@@ -213,7 +204,7 @@ export class ConvivaAnalytics {
       return;
     }
 
-    this.client.sendCustomEvent(this.sessionKey, eventName, eventAttributes);
+    // TODO: Look further into custom events. The API from 2.x -> 4.x appears to have lost the option for custom events
```
mckennajones: Might be nice to log out a debug warning message here to alert integrators. Or at the very least this should be called out in the Changelog.

src/ts/ConvivaAnalytics.ts
```diff
@@ -655,12 +650,20 @@ export class ConvivaAnalytics {
     this.onPlaybackStateChanged(event);
   };
 
+  private onTimeChanged = (event: TimeChangedEvent) => {
+    if (!this.isSessionActive()) {
+      return;
+    }
+
+    this.playheadTimeMs = event.time * 1000;
```
mckennajones: Does Conviva provide any guidance on the expected value for Live streams? Asking because `event.time` will be the absolute epoch time in live streams. We could also provide the relative time if that is required (`player.getCurrentTime('relativetime')`)

mckennajones (commented):
Changes look good overall, just a couple minor comments. Agree with Daniel that a PR description would be useful.

src/ts/ConvivaAnalytics.ts
gregdolby: My understanding is Conviva expects relative time. This way seems better than using event.time so I'll switch to it. Does player 8.6.0 support (player.getCurrentTime('relativetime'))? Because the player api typings file is telling me it doesn't expect any arguments. Or does this always return relative time without an argument? Even for live?

dweinber: No, 8.6.0 doesn't support this, it was added in [8.31.0](https://bitmovin.com/docs/player/releases/web/web-8-31-0). We may want to raise the minimum player version.

dweinber:
Looks like `package-lock.json` isn't in sync with the `package.json`: https://travis-ci.org/github/bitmovin/bitmovin-player-analytics-conviva/builds/753989259#L209-L221 @turbidwater. Likely due to the player version update. Running `npm i` and committing the lock file should do the trick.

dweinber (approved):
Looks good to me know! Thanks for addressing all the requests!

mckennajones (approved):
Thanks for addressing my points. Looks good to me now.

turbidwater:
Thanks! We are ready for you guys merge it when you are ready.

PR #67, opened 2021-01-25
fix: Remove redundant Conviva release calls in destroy method

Thomasvdam:
Both these resources are alreadt taken care of by the internalEndSession
call in the line above.


dweinber (approved):
Looks good and makes sense! Thanks for contributing!
I'll take care of adding an entry to the CHANGELOG.md file.

PR #68, opened 2021-02-22
Removed clearing of metadata at playback completion

turbidwater:
We removed content metadata reset from playback finished event. It will still reset out on source unload or when session end event is called. An alternative fix would be to call conviva.updateContentMetadata on the conviva object returned from the bitmovin player when a replay is replayed. Similar to how it is done in the index.html file initially.

ex:
conviva.updateContentMetadata({ applicationName: 'Bitmovin Player Conviva Analytics Integration Test Page', viewerId: 'uniqueViewerId', custom: { appVersion: '1.0', contentId: 'someContentId', playerVendor: 'bitmovin', playerVersion: player.version }, });

### DIFF
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -555,7 +555,6 @@ export class ConvivaAnalytics {
     this.onPlaybackStateChanged(event);
     this.convivaVideoAnalytics.release();
     this.convivaVideoAnalytics = null;
-    this.resetContentMetadata();
   };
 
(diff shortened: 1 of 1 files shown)


dweinber (approved):
This looks almost too easy 😅  But makes sense 👍

dweinber:
Looks like you need to merge `bitmovin:develop` into your branch first @turbidwater. And please add a changelog entry (missed that in the review)

aneurinc (approved):

turbidwater:
@dweinber Updated with bitmovin:develop

PR #61, opened 2020-07-09
Make possible to override Html5Metadata values (Solves #60)

cristian-atehortua:
Solves #60

Issue #60, opened 2020-07-09
Make possible to override Html5Metadata default values


cristian-atehortua:
In some cases, a user of this library could want to override the default values generated by Conviva in MetadataInterface. For example, if you are using Bitmovin in a LG Smart TV, you want to send *webOS* as "Operating System" and its version.

Currently, the only field that a user is able to override is "Device Category", can this be done for other values in Html5Interface?

It would be nice to have this feature for our integration of this Library in our LG Smart TV app.

Additional fields I think it would be nice to be overridden:
- Device Brand
- Device Manufacturer
- Device Model
- Device Type
- Device Version
- Operating System Name
- Operating System Version

dweinber:
Implemented in https://github.com/bitmovin/bitmovin-player-analytics-conviva/pull/72 and released in [v4.0.3](https://github.com/bitmovin/bitmovin-player-analytics-conviva/releases/tag/v4.0.3)

PR #72, opened 2021-04-16
Added configuration option to set Conviva Device Metadata

cristian-atehortua:
Solves #60.

Same as #61 but created from version 4.0.2


package.json
```diff
@@ -37,5 +37,8 @@
     "webpack-cli": "^3.1.2",
     "webpack-dev-server": "^3.10.3",
     "webpack-merge": "^4.1.5"
+  },
+  "peerDependencies": {
+    "bitmovin-player": "^8.31.0"
```
dweinber, line 42: I think we don't need the `bitmovin-player` in both, `devDependencies` and `peerDependencies`, right? I think having it in `peerDependencies` makes sense, so I'd suggestion removing it from `devDependencies` if there are no good reasons to keep it there.

src/ts/Conviva.d.ts
```diff
@@ -146,10 +146,14 @@ declare namespace Conviva {
     }
   }
 
+  type DeviceMetadata = {
```
dweinber, line 149: Why do we redefine a rather weak type here (Why do we need the `undefined`, btw - should be avoided in types unless really necessary)? We have this properly defined in an interface in `ConvivaAnalytics.ts`. If we'd use that, the user would have a better type checking and potentially auto-completion to create the metadata object.

src/ts/ConvivaAnalytics.ts
```diff
@@ -113,8 +166,16 @@ export class ConvivaAnalytics {
     this.sessionKey = Conviva.Constants.NO_SESSION_KEY;
     this.isAd = false;
 
-    const deviceMetadata: {[key: string]: Conviva.Constants.DeviceCategory} = {
-      [Conviva.Constants.DeviceMetadata.CATEGORY]: this.config.deviceCategory,
+    const deviceMetadataFromConfig = this.config.deviceMetadata || {};
+    const deviceMetadata = {
```
dweinber, line 170: Is it recommended by Conviva to always send all of the `deviceMetadata` fields, even if they might be `undefined`?

dweinber (changes requested):

package.json
cristian-atehortua: `peerDependencies` are not installed when you run `npm install`. They are useful to tell the consumers what dependencies your package need to be able to run. It is needed to keep `bitmovin-player` in `devDependencies` to be able to import the types from that package.

src/ts/ConvivaAnalytics.ts
cristian-atehortua: In practice, for Conviva library, sending them as undefined is the same that not sending them at all. We need to define in this way because the user can provide those values or not.

src/ts/Conviva.d.ts
cristian-atehortua: If you see where this type is used and the comparison with previous changes you will see that they are pretty similar. The difference is supporting other types for the values of this object because with new device metadata options can have other types.
Regarding the use of `undefined` it could be deleted.

cristian-atehortua:
@dweinber Hi, I answered your comments and updated the PR with a change related to them, can you review it again?

src/ts/ConvivaAnalytics.ts
```diff
@@ -34,6 +32,61 @@ export interface ConvivaAnalyticsConfiguration {
    * user agent string parsing by the Conviva SDK. (default: WEB)
    */
   deviceCategory?: Conviva.Constants.DeviceCategory;
```
stonko1994, line 35: I think we should deprecate this property here in favor our the `deviceMetadata.category`

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- `bitmovin-player@^8.31.0` as peer dependency.
+- Configuration option to set Conviva Device Metadata
```
stonko1994:
```suggestion
- Possibility to set detailed Conviva Device Metadata via `ConvivaAnalyticsConfiguration.deviceMetadata`
```

If we going to deprecate the existing one please add a `### Deprecated` section.

stonko1994 (commented):
@dweinber what do you think about deprecating the existing property? 🤔

src/ts/ConvivaAnalytics.ts
cristian-atehortua: Hi, @stonko1994. Please confirm if should I deprecate that field and what is the process you have to do it. (Just adding a `@deprecated` JSDoc tag? Adding a warning to the console when this field is used?)

cristian-atehortua:
@stonko1994 @dweinber Can you please check out the comments and answer them? Thanks

stonko1994 (commented):
From my point of view, we should deprecate it however we need the input from @dweinber as well.

src/ts/ConvivaAnalytics.ts
stonko1994: Yes, just a `@deprecated` JSDoc tag is enough 🙁

package.json
dweinber: Ah right 👍  Thanks

Shouldn't we use the same version actually in the `devDependencies` (i.e. adding `^`)?

src/ts/Conviva.d.ts
dweinber: Having spent a bit of time on this now I agree, let's keep it as it is 😅

src/ts/ConvivaAnalytics.ts
dweinber: I see, thanks for explaining. Let's keep it this way, keeps the implementation simple 👍

dweinber (approved):
Looks good now 👍  
Please just have a look at my comment in `package.json`.

@stonko1994 good idea to deprecate the old field, thanks.

package.json
cristian-atehortua: Yes, I will update it.

cristian-atehortua:
@dweinber I just updated with your last comment.

Thank you both @stonko1994 and @dweinber  for your help. If you don't have any other comments you can merge when you want.

dweinber:
Great, thanks!
I'll just wait for @stonko1994's approval (or change request) and then we're good to merge.

stonko1994 (approved):
LGTM 👍 Feel free to merge 🙂

cristian-atehortua:
@dweinber @stonko1994 Thanks again.
I think one of you should merge cause "Only those with write access to this repository can merge pull requests."

image

PR #74, opened 2021-04-29
Adding possibility for application to override internally used custom metadata tags

luckygoyal-bitmovin:
Currently `streamType`, `playerType` and `vrContentType` are 3 tags which are defined and populated internally by the Conviva integration. This PR is to allow application to override the values set in these internally.
- Customers may already be using these custom tag names but have a different set of values for these. So they would want to override the internal values.
- Allowing to override should not result in any issue as the values will be overridden if explicitly set externally. If customer want to use internal values, they should not override these. So allowing the override makes it possible to handle both scenarios.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - `bitmovin-player@^8.31.0` as peer dependency.
 - Possibility to set detailed Conviva Device Metadata via `ConvivaAnalyticsConfiguration.deviceMetadata`
-- Possibility to override values for internally set custom metdata keys streamType, playerType and vrContentType
+- Possibility to override values for internally set custom metdata keys `streamType`, `playerType` and `vrContentType`
 
 ### Deprecated
 - `ConvivaAnalyticsConfiguration.deviceCategory` field in favour of `ConvivaAnalyticsConfiguration.deviceMetadata.category` field
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - `bitmovin-player@^8.31.0` as peer dependency.
 - Possibility to set detailed Conviva Device Metadata via `ConvivaAnalyticsConfiguration.deviceMetadata`
+- Possibility to override values for internally set custom metdata keys streamType, playerType and vrContentType
 
 ### Deprecated
 - `ConvivaAnalyticsConfiguration.deviceCategory` field in favour of `ConvivaAnalyticsConfiguration.deviceMetadata.category` field
--- a/spec/tests/ContentMetadata.spec.ts
+++ b/spec/tests/ContentMetadata.spec.ts
@@ -1,5 +1,5 @@
 /// <reference path='../../src/ts/Conviva.d.ts'/>
-import { PlayerType } from 'bitmovin-player';
+import {PlayerType, StreamType, VRContentType} from 'bitmovin-player';
 import { ConvivaAnalytics } from '../../src/ts';
 import { MockHelper, TestingPlayerAPI } from '../helper/MockHelper';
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - `bitmovin-player@^8.31.0` as peer dependency.
 - Possibility to set detailed Conviva Device Metadata via `ConvivaAnalyticsConfiguration.deviceMetadata`
+- Possibility to override values for internally set custom metdata keys streamType, playerType and vrContentType
```
dweinber: I'd suggest putting code in backticks to make it obvious that those are the actual names from the code:
```suggestion
- Possibility to override values for internally set custom metdata keys `streamType`, `playerType` and `vrContentType`
```

spec/tests/ContentMetadata.spec.ts
```diff
@@ -1,5 +1,5 @@
 /// <reference path='../../src/ts/Conviva.d.ts'/>
-import { PlayerType } from 'bitmovin-player';
+import {PlayerType, StreamType, VRContentType} from 'bitmovin-player';
```
dweinber: Please stick to the existing code style 🙂 
```suggestion
import { PlayerType, StreamType, VRContentType } from 'bitmovin-player';
```

spec/tests/ContentMetadata.spec.ts
```diff
@@ -43,6 +43,54 @@ describe('content metadata spec', () => {
...
+
+      expect(convivaVideoAnalytics.reportPlaybackRequested).toHaveBeenLastCalledWith(expect.objectContaining({
+        vrContentType: undefined,
+      }));
+    });
+
+    it('override playerType in custom tags', () => {
+      jest.spyOn(playerMock, 'getPlayerType').mockReturnValue(PlayerType.Native);
+      jest.spyOn(playerMock, 'getStreamType').mockReturnValue(StreamType.Dash);
+      convivaAnalytics.updateContentMetadata({ custom: { myTag: 'withMyValue', playerType: PlayerType.Html5 }, assetName: 'MyAsset' });
```
dweinber: Nit: Not 100% related to this PR, but might be worth adding a test case where we check that `myTag` also ends up in the object `reportPlaybackRequested` is called with.

dweinber (changes requested):
Nice and simple solution - well done 👍 
Just some small improvements to be done, I think we're good to merge after those.

spec/tests/ContentMetadata.spec.ts
```diff
@@ -34,7 +34,7 @@ describe('content metadata spec', () => {
       }));
     });
 
-    it('set player type in custom tags', () => {
+    it('default playerType in custom tags', () => {
```
stonko1994: Those tests names should be written like a sentence:
- `it sets the default playerType in custom tags` 
vs
- `it default playerType in custom tags`

```suggestion
    it('sets the default playerType in custom tags', () => {
```

spec/tests/ContentMetadata.spec.ts
```diff
@@ -43,6 +43,54 @@ describe('content metadata spec', () => {
       }));
     });
 
+    it('default streamType in custom tags', () => {
```
stonko1994:
```suggestion
    it('sets the default streamType in custom tags', () => {
```

spec/tests/ContentMetadata.spec.ts
```diff
@@ -43,6 +43,54 @@ describe('content metadata spec', () => {
...
+      jest.spyOn(playerMock, 'getStreamType').mockReturnValue(StreamType.Dash);
+      playerMock.eventEmitter.firePlayEvent();
+
+      expect(convivaVideoAnalytics.reportPlaybackRequested).toHaveBeenLastCalledWith(expect.objectContaining({
+        streamType: 'dash',
+      }));
+    });
+
+    it('default vrContentType in custom tags', () => {
```
stonko1994:
```suggestion
    it('sets the default vrContentType in custom tags', () => {
```

stonko1994 (approved):
Minor improvement to the tests but approving already 🙂

spec/tests/ContentMetadata.spec.ts
luckygoyal-bitmovin: A test case for custom tags already exists.

```
describe('externally content metadata updates', () => {
   ......................................................
   ..................................................... 
   it('custom tags', () => {
```

CHANGELOG.md
luckygoyal-bitmovin: Done.

spec/tests/ContentMetadata.spec.ts
luckygoyal-bitmovin: Done

spec/tests/ContentMetadata.spec.ts
luckygoyal-bitmovin: Done

spec/tests/ContentMetadata.spec.ts
luckygoyal-bitmovin: Done

spec/tests/ContentMetadata.spec.ts
luckygoyal-bitmovin: Done

spec/tests/ContentMetadata.spec.ts
dweinber: I searched for this but still missed it 🤦 Thanks for pointing me to it!

dweinber (approved):
Looks good, thanks! 👍

thomhs22 (approved):
looking good to me.

PR #77, opened 2021-06-08
feat: Add reporting rendered framerate

realeyes-craig:
Added:
- Reporting framerate if available on `targetQuality`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- Tracking the current framerate
 
 ## [4.0.3]
 ### Added
--- a/spec/helper/MockHelper.ts
+++ b/spec/helper/MockHelper.ts
@@ -241,7 +241,7 @@ interface EventEmitter {
 
   fireAdErrorEvent(): void;
 
-  fireVideoPlaybackQualityChangedEvent(bitrate: number): void;
+  fireVideoPlaybackQualityChangedEvent(bitrate: number, frameRate: number): void;
 
   fireCastStartedEvent(resuming?: boolean): void;
 
@@ -422,7 +422,7 @@ class PlayerEventHelper implements EventEmitter {
     });
   }
 
-  fireVideoPlaybackQualityChangedEvent(bitrate: number): void {
+  fireVideoPlaybackQualityChangedEvent(bitrate: number, frameRate: number): void {
     this.fireEvent<VideoPlaybackQualityChangedEvent>({
       timestamp: Date.now(),
       type: PlayerEvent.VideoPlaybackQualityChanged,
@@ -437,6 +437,7 @@ class PlayerEventHelper implements EventEmitter {
         bitrate: bitrate,
         width: null,
         height: null,
+        frameRate: frameRate,
       },
     });
   }
--- a/spec/tests/PlayerEvents.spec.ts
+++ b/spec/tests/PlayerEvents.spec.ts
@@ -248,18 +248,26 @@ describe('player event tests', () => {
     });
   });
 
-  describe('bitrate tracking', () => {
+  describe('video quality tracking', () => {
     it('report bitrate on event', () => {
       playerMock.eventEmitter.firePlayEvent();
       playerMock.eventEmitter.firePlayingEvent();
-      playerMock.eventEmitter.fireVideoPlaybackQualityChangedEvent(2_400_000);
+      playerMock.eventEmitter.fireVideoPlaybackQualityChangedEvent(2_400_000, 30);
 
       expect(convivaVideoAnalytics.reportPlaybackMetric).toHaveBeenCalledWith(Conviva.Constants.Playback.BITRATE, 2_400);
     });
 
+    it('report framerate on event', () => {
+      playerMock.eventEmitter.firePlayEvent();
+      playerMock.eventEmitter.firePlayingEvent();
+      playerMock.eventEmitter.fireVideoPlaybackQualityChangedEvent(2_400_000, 30);
+
+      expect(convivaVideoAnalytics.reportPlaybackMetric).toHaveBeenCalledWith(Conviva.Constants.Playback.RENDERED_FRAMERATE, 30);
+    });
+
     describe('event order workaround', () => {
       it('track current bitrate on session initialization', () => {
-        playerMock.eventEmitter.fireVideoPlaybackQualityChangedEvent(4_800_000);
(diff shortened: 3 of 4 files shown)


src/ts/ConvivaAnalytics.ts
```diff
@@ -631,6 +631,9 @@ export class ConvivaAnalytics {
```
stonko1994, line 631: There is a comment in line 409 stating that the framerate is unused. Please also update this comment 🙂

src/ts/ConvivaAnalytics.ts
```diff
@@ -631,6 +631,9 @@ export class ConvivaAnalytics {
 
     this.lastSeenBitrate = null;
     this.convivaVideoAnalytics.reportPlaybackMetric(Conviva.Constants.Playback.BITRATE, bitrateKbps);
+    if (event.targetQuality.frameRate) {
+      this.convivaVideoAnalytics.reportPlaybackMetric(Conviva.Constants.Playback.RENDERED_FRAMERATE, event.targetQuality.frameRate);
+    }
```
stonko1994, line 636: As this is called inside the `onVideoQualityChanged` it could happen that `event.targetQuality.frameRate` changed from the previous quality.

When we just check `if (event.targetQuality.frameRate)` but the `frameRate` changes to `undefined / null` we would still report the previous framerate 🤔 Is there a way to unset the `Conviva.Constants.Playback.RENDERED_FRAMERATE` if it's not set?

stonko1994 (commented):
- Please add a Changelog entry
- Please open the PR against `develop`
- Optional: Please add unit tests 🙂

src/ts/ConvivaAnalytics.ts
realeyes-craig: That line seems to be referring to encoded framerate. This PR is for reporting rendered framerate which calculates "Average FPS" for a session.
It seems like this metric does not fit into the same category being defined in that comment.

src/ts/ConvivaAnalytics.ts
realeyes-craig: I do not see a way to _unset_ this metric. The intention is for this metric to change and not report it if we see `undefined / null`.
What is the situation where we would report a previous framerate?

src/ts/ConvivaAnalytics.ts
stonko1994: ah oke 👍

src/ts/ConvivaAnalytics.ts
stonko1994: I mean like different qualities could have different / no framerates e.g. (don't know 100%) switching to an audio-only quality could mean that there maybe is no framerate present 🤔 

@dweinber maybe has more insights into how the web-player behaves here

realeyes-craig: It's my understanding that if the qualities have different framerate that should be reported to calculate the average FPS over a session.
I'm unsure on how no framerate affects the end calculations in conviva and don't see anything in their documentation for handling this case.

stonko1994: Ok 🤔 Will approve the PR 🙂

stonko1994 (approved):

src/ts/ConvivaAnalytics.ts
dweinber: Switching between audio/video and audio-only renditions is anyway not supported in the web player @stonko1994 (except Safari) 😅 

Framerate is optional, so theoretically, one rendition/Representation could have it set while another has not. However, I can't remember having seen this in practice, it's usually all or nothing. So should be fine.

dweinber:
Just to be clear on `rendered framerate` - what the player provides is the framerate specified in the manifest for that video quality. Is this the intention of `RENDERED_FRAMERATE` or is this another field @realeyes-craig?

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- Tracking the current framerate
```
dweinber, line 9:
```suggestion
- Tracking the framerate for the currently played video quality based on the framerate specified in the manifest file
```

src/ts/ConvivaAnalytics.ts
stonko1994: Ok, thx for clarifying 🙂

spec/helper/MockHelper.ts
```diff
@@ -241,7 +241,7 @@ interface EventEmitter {
 
   fireAdErrorEvent(): void;
 
-  fireVideoPlaybackQualityChangedEvent(bitrate: number): void;
+  fireVideoPlaybackQualityChangedEvent(bitrate: number, frameRate: number): void;
```
dweinber, line 244: Please make `frameRate` optional, so we can test events with and without `frameRate` being available.

spec/helper/MockHelper.ts
```diff
@@ -422,7 +422,7 @@ class PlayerEventHelper implements EventEmitter {
     });
   }
 
-  fireVideoPlaybackQualityChangedEvent(bitrate: number): void {
+  fireVideoPlaybackQualityChangedEvent(bitrate: number, frameRate: number): void {
```
dweinber, line 425: Please make `frameRate` optional, so we can test events with and without `frameRate` being available.

spec/helper/MockHelper.ts
```diff
@@ -437,6 +437,7 @@ class PlayerEventHelper implements EventEmitter {
         bitrate: bitrate,
         width: null,
         height: null,
+        frameRate: frameRate,
```
dweinber, line 440: Please only expose `frameRate` if it's passed in the signature

spec/tests/PlayerEvents.spec.ts
```diff
@@ -248,18 +248,26 @@ describe('player event tests', () => {
     });
   });
 
-  describe('bitrate tracking', () => {
+  describe('video quality tracking', () => {
     it('report bitrate on event', () => {
       playerMock.eventEmitter.firePlayEvent();
       playerMock.eventEmitter.firePlayingEvent();
-      playerMock.eventEmitter.fireVideoPlaybackQualityChangedEvent(2_400_000);
+      playerMock.eventEmitter.fireVideoPlaybackQualityChangedEvent(2_400_000, 30);
```
dweinber, line 255: Please add a test case where we don't pass a frame rate and explicitly check that we don't set it (so we don't accidentally break it in the future)

dweinber (changes requested):
Please clarify if `RENDERED_FRAMERATE` is the right field for the data we provide.

Code change looks good and straightforward, please see the comments around the tests.

PR #85, opened 2022-10-05
Support consecutive playback

### LINKED ISSUE #76: When loading a new source without releasing the instance, there is no new session created
There is a bug with version 4.0.3 (and probably 4.x.x)

When the player is reproducing a source and the user loads a new one without releasing the Conviva Analytics instance, the previous session is stopped but the new session with the new source is not created and reported. This behavior is different from version 3.

The error I can see in the console is:
```
unhandled exception in event callback Error: Constants not yet configured
    at Object.buildVideoAnalytics (conviva-core-sdk.js:211)
    at e.internalInitializeSession (vendors~main.9bb4ed95.chunk.js:2)
    at e.onPlay (vendors~main.9bb4ed95.chunk.js:2)
```

martinstark:
Close  #76

Releasing the Conviva analytics class makes it unusuable without re-initializing it from the ground up, this is a minimal fix in order to support consecutive playback using BMPlayer with consecutive .load() - .unload() or .load() - .load() patterns.

There is still the concern that the Conviva docs seem to indicate that a single convivaVideoAnalytics instance should be used for the lifetime of the player instance, but in the interest of time I've opted to not rewrite this.

The fix has only been lightly tested using the conviva Touchstone where two consecutive .load() calls to the player will close and initiate new session as expected. I have not run the full validation suite (again, in the interest of time).


example/index.html
```diff
@@ -40,41 +41,94 @@
...
+        });
+
+        console.log('Successfully created Bitmovin Player instance');
+      }, function(reason) {
+        console.error('Error while creating Bitmovin Player instance');
         console.error(reason);
       });
+    }
+
+    loadPlayer();
```
dweinber: You are now calling `loadPlayer` right after `createPlayer` although `createPlayer` is now an `async` function. I think we may want to wait until `createPlayer` finished, right?

martinstark: Good point, createPlayer wasn't async at first but realized I have to await the if-wrapped .destroy-call before creating a new player. I could also potentially remove the destroy call from the create method and require the user to manually press the destroy button before clicking create player again. It's just there for convenience.

It still seems to work, but yes it should be awaited.

Before I fix it, I also realize I probably shouldn't be using async/await at all if we want the demo setup to be easily usable without transpilation (it's avoiding arrow functions and let/const for this reason). Should I convert it to regular promise syntax?

dweinber: Yes please, let's avoid async/await in the demo. 
I think it's fine if the user has to use destroy explicitly in the demo.

Just FYI: most things in `destroy` are in fact synchronous, so it would mostly work; However, it's not guaranteed and as soon as you use DRM things might break as clearing the DRM sessions with the browser could take a bit.

martinstark: Sync destroy->create doesn't work if destroy isn't awaited, at least not the way the demo app is structured with re-using container element. It ends up in a broken layout state.

I'll remove the convenience-destroy inside createPlayer and add an ugly "alert" that you need to destroy the player before creating it again.

dweinber (approved):

PR #99, opened 2023-08-22
Support for new metrics Audio, Subtitle & Closed Caption language

glaucocodes:
### Added
* Support for Conviva-SDK version 4.7.0
* Tracking for Audio language
* Tracking for Subtitle Language
* Tracking for Closed Caption Language

### Removed
* Support for Conviva-SDK version 4.6.1


src/ts/ConvivaAnalytics.ts
```diff
@@ -463,6 +467,19 @@ export class ConvivaAnalytics {
     if (this.lastSeenBitrate) {
       this.convivaVideoAnalytics.reportPlaybackMetric(Conviva.Constants.Playback.BITRATE, this.lastSeenBitrate);
     }
+
+    // Send the session init audio language values.
+    this.updateAudioTrack(this.player.getAudio());
+
+    // Check if at session init has a subtitle enabled.
+    const enableSubtitle = this.player.subtitles.list().filter((i) => i.enabled);
```
dweinber: Please note that `this.player.subtitles` may be undefined if no subtitle modules are added to the modular player. Please make sure to cover this case as otherwise the code would run into an exception in this scenario.

dweinber (changes requested):
Thanks for submitting this PR with some nice features along with some tests for them!

There's one thing I spotted which needs to be slightly reworked to avoid a run-time exception but should be quick. The rest looks good, tests succeed and these changes can (and will) be merged once that one requested change is done

src/ts/ConvivaAnalytics.ts
glaucocodes: Thanks for your feedback, I just push commit with the fix.

dweinber (approved):
Looks good, thanks for addressing the feedback, and thanks again for your contribution!

PR #104, opened 2024-05-14
Basic client ad tracking using `Conviva.AdAnalytics`

karpov-kir:
Related to: https://bitmovin.atlassian.net/browse/TA-2596.

- Add client-side ad tracking via `AdAnalytics` (the rest is reported via `VideoAnalytics`)
- Remove some redundant fixes that were marked with TODO since the order of events had been fixed according to my tests

I suspect that the implementation is not 100% complete, but it works and reports ads at least in its basic form. I think it makes sense to merge it at this stage and submit it for @saravanans-github review.

image

PR #105, opened 2024-05-14
Migrate to latest Conviva types

karpov-kir:
Related to: https://bitmovin.atlassian.net/browse/TA-2596

- Use the latest Conviva types
- Explicitly add `./@types/conviva/ConvivaExtension.d.ts` with types that are missing in the core package types
- Update `ts-loader` and TypeScript so that the project compiles fine (I've added some comments in the PR with explanations)

This PR is required to finalize https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/104.


package.json
```diff
@@ -40,9 +40,9 @@
     "prettier": "^2.8.8",
     "string-replace-loader": "^2.1.1",
     "ts-jest": "^23.10.5",
-    "ts-loader": "^5.3.1",
+    "ts-loader": "^8.4.0",
```
karpov-kir, line 43: The previous version cannot load `./@types/conviva/*.d.ts` files, but it works fine in 8.x.

package.json
```diff
@@ -40,9 +40,9 @@
...
     "tslint": "^5.11.0",
-    "typescript": "^3.2.2",
+    "typescript": "^4.9.5",
```
karpov-kir, line 45: This is required to make this types utility https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/105/files#diff-26e5b6095fb9edb40590c6bc9eba1c9eb6fbefe6269e3a493656fa85bef63b79R1 work. 

The following syntax is not supported in 3.x `K in keyof T as string`.

spec/helper/MockHelper.ts
```diff
@@ -359,8 +358,8 @@ class PlayerEventHelper implements EventEmitter {
       type: PlayerEvent.AdSkipped,
       ad: {
         isLinear: true,
-        width: null,
-        height: null,
+        width: 0,
```
karpov-kir, line 361: Just make it types-compatible (it cannot be `null` according to the latest types).

spec/tests/PlayerEvents.spec.ts
```diff
@@ -242,7 +240,6 @@ describe('player event tests', () => {
       expect(convivaVideoAnalytics.reportAdBreakStarted).toHaveBeenCalledWith(
         Conviva.Constants.AdType.CLIENT_SIDE,
         Conviva.Constants.AdPlayer.SEPARATE,
-        Conviva.Constants.AdPosition.PREROLL,
```
karpov-kir, line 245: Does not exist in the latest types.

src/ts/ContentMetadataBuilder.ts
```diff
@@ -83,20 +81,26 @@ export class ContentMetadataBuilder {
...
+      [Conviva.Constants.DURATION]: this.contentMetadata.duration,
+      [Conviva.Constants.DEFAULT_RESOURCE]: this.contentMetadata.defaultResource,
+      [Conviva.Constants.STREAM_URL]: this.contentMetadata.streamUrl,
+      [Conviva.Constants.IS_LIVE]: this.contentMetadata.streamType,
+      [Conviva.Constants.VIEWER_ID]: this.contentMetadata.viewerId || 'GET_VIEWER_ID_FROM_PLAYER',
+      [Conviva.Constants.PLAYER_NAME]: this.contentMetadata.applicationName || 'GET_PLAYER_NAME_OR_TYPE',
+    };
+    const customContentInfo: CustomContentMetadata = {
+      [Conviva.Constants.FRAMEWORK_NAME]: this.contentMetadata.framework,
+      [Conviva.Constants.FRAMEWORK_VERSION]: this.contentMetadata.frameworkVersion,
```
karpov-kir: These two are not present in `ReservedContentMetadata` props, so they should be custom data.

src/ts/ConvivaAnalytics.ts
```diff
@@ -354,7 +354,6 @@ export class ConvivaAnalytics {
     this.convivaVideoAnalytics.reportAdBreakStarted(
       Conviva.Constants.AdType.CLIENT_SIDE,
       Conviva.Constants.AdPlayer.SEPARATE,
-      Conviva.Constants.AdPosition.PREROLL,
```
karpov-kir, line 357: Does not exist in the latest types.

tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
```
karpov-kir, line 1: Split into `tsconfig.json` and `tsconfig.build.json` so that the `spec` and the `src` folders are not included in the `dist` folder. We cannot use:

```
"include": [
    "./src"
  ]
```

because tests then don't have access to `Conviva.d.ts` types. If we remove `include` completely, then redundant files are built into the `dist` folder.

image

it needs to be:

image

src/ts/ContentMetadataBuilder.ts
dweinber: I think it would make sense to deprecate/remove those two options in that case. Customers can set any custom data anyway. What do you think?

dweinber (commented):
Code looks good to me. Just one open question about the now-custom framework fields

src/ts/ContentMetadataBuilder.ts
karpov-kir: Done in https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/105/commits/04a0c70fd2c0e4686975bbefd62ad16af267adc0, https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/105/commits/529eb4d0064f1118cfa21efa8e9ea2b75f9ab5a9.

dweinber (approved):

PR #106, opened 2024-05-16
Leverage Conviva types

karpov-kir:
- Install `@convivainc/conviva-js-coresdk` as a peer dependency and reference types in the code explicitly
- Remove redundant types
- Properly reference Conviva types in the source code and use `Conviva.ContentMetadata` as a basis instead custom re-declaration of the whole `Metadata`
- Make the metadata `build` more explicitly follow the types (just split it into methods that match types semantically)

The above is done to reduce confusion and make future extensions easier to follow.


webpack.prod.js
```diff
@@ -2,5 +2,5 @@ const merge = require('webpack-merge');
 const common = require('./webpack.common.js');
 
 module.exports = merge(common, {
-  mode: 'production',
+  mode: 'development',
```
dweinber: I don't think we should change the mode to `development` in the `webpack.prod.js` 🤔

example/index.html
```diff
@@ -74,8 +74,6 @@
         conviva.updateContentMetadata({
           applicationName: 'Bitmovin Player Conviva Analytics Integration Test Page',
           viewerId: 'uniqueViewerId',
-          framework: 'Bitmovin Player',
-          frameworkVersion: player.version,
```
karpov-kir, line 78: Forgot to remove them in a previous PR. They don't exist here anymore.

src/ts/ContentMetadataBuilder.ts
```diff
@@ -1,47 +1,30 @@
...
-
-  // Dynamic
-  encodedFrameRate?: number;
-  defaultResource?: string;
-  streamUrl?: string;
-  framework?: string;
-  frameworkVersion?: string;
-}
+export type Metadata = Conviva.ContentMetadata & {
+  // TODO extend it with additional standard tags
```
karpov-kir, line 8: It will be done in a follow-up PR.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Deprecated
 - Removed `framework` and `frameworkVersion` custom metadata fields (custom tags)
+- Removed `CustomContentMetadata`, use `Metadata['custom']` instead
```
dweinber: Shouldn't these actually be in a `### Removed` section?

src/ts/ContentMetadataBuilder.ts
karpov-kir: Done in https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/108.

src/ts/ContentMetadataBuilder.ts
```diff
@@ -52,47 +35,72 @@ export class ContentMetadataBuilder {
...
-      this.contentMetadata.duration = this.metadataOverrides.duration || this.metadata.duration;
+      metadata.viewerId = this.viewerId;
+      metadata.streamType = this.metadataOverrides.streamType || this.metadata.streamType;
+      metadata.applicationName = this.metadataOverrides.applicationName || this.metadata.applicationName;
+      metadata.duration = this.metadataOverrides.duration || this.metadata.duration;
 
-      this.contentMetadata.custom = this.custom;
+      metadata.custom = this.custom;
     }
+    // If the playback has been started, the values cannot be changed and the latest values before the playback started has to be used
```
dweinber: Comment should be moved into the `else` block

src/ts/ContentMetadataBuilder.ts
```diff
@@ -52,47 +35,72 @@ export class ContentMetadataBuilder {
...
+
+    const newReservedMetadata: ReservedContentMetadata = {
+      [Conviva.Constants.ASSET_NAME]: newMetadata.assetName,
+      [Conviva.Constants.ENCODED_FRAMERATE]: newMetadata.encodedFrameRate,
+      [Conviva.Constants.DURATION]: newMetadata.duration,
+      [Conviva.Constants.DEFAULT_RESOURCE]: newMetadata.defaultResource,
+      [Conviva.Constants.STREAM_URL]: newMetadata.streamUrl,
+      [Conviva.Constants.IS_LIVE]: newMetadata.streamType,
+      [Conviva.Constants.VIEWER_ID]: newMetadata.viewerId || 'GET_VIEWER_ID_FROM_PLAYER',
+      [Conviva.Constants.PLAYER_NAME]: newMetadata.applicationName || 'GET_PLAYER_NAME_OR_TYPE',
```
dweinber, line 102: Looks like it was like that already, but matching `applicationName` to `PLAYER_NAME` sounds a bit weird 🤔

src/ts/ContentMetadataBuilder.ts
```diff
@@ -121,11 +129,11 @@ export class ContentMetadataBuilder {
     this.metadata.applicationName = newValue;
   }
 
-  set custom(newValue: CustomContentMetadata) {
+  set custom(newValue: Metadata['custom']) {
```
dweinber, line 136: Didn't know you can use types like that 😲 Thanks for showing that to me!

src/ts/index.ts
```diff
@@ -1,2 +1,5 @@
+// Import to extend Conviva types.
+import './conviva/ConvivaExtension';
```
dweinber, line 2: Would it make sense to also export the Conviva objects? Because currently I believe constants like `Conviva.Constants.DeviceCategory.WEB` (which we also show in the README) might not be accessible (or are they?).

dweinber (changes requested):
Nice update! A few comments though

webpack.prod.js
karpov-kir: Hmm, this is not in this branch. I did it for testing but then cleaned it up, this has to be `production`. You somehow looked at an outdated version. Could you take another look?

src/ts/ContentMetadataBuilder.ts
karpov-kir: Do you mean this way? https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/106/commits/70cad66a428f20a7bfbb2d0616c9387fe878b848

src/ts/ContentMetadataBuilder.ts
karpov-kir: Yeah, it's weird. I commented on it in the sample https://github.com/bitmovin/bitmovin-player-web-samples/blob/main/react/bundled-player/package-based-with-conviva/src/App.tsx#L77.

I've just added a comment in this repo as well - https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/106/commits/c560691e614fdd9a01041493253c90ad67d95b83.

karpov-kir: And here is one more comment https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/106/commits/06cd7394d6fe438d8208460283162123d847bf9f.

src/ts/ContentMetadataBuilder.ts
karpov-kir: Yeah, it's weird. I commented on it in the sample https://github.com/bitmovin/bitmovin-player-web-samples/blob/main/react/bundled-player/package-based-with-conviva/src/App.tsx#L77.

I've just added a comment in this repo as well - https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/106/commits/c560691e614fdd9a01041493253c90ad67d95b83.

karpov-kir: And here is one more comment https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/106/commits/06cd7394d6fe438d8208460283162123d847bf9f.

src/ts/index.ts
karpov-kir: Users can import constants from the Conviva package:

image

karpov-kir: Once there is an import from `@bitmovin/player-integration-conviva` the extension is imported automatically as well:

image

The `StorageInterface` is from `ConvivaExtension`. So if I CMD + click on the `StorageInterface` it is resolved to:

image

NOTE: I have mapped `@bitmovin/player-integration-conviva` to a local folder for development:

image

CHANGELOG.md
karpov-kir: Done in https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/106/commits/cce84d26764897d099617e79b7bee9d625809b41.

webpack.prod.js
dweinber: Looks like that file isn't in the changed ones anymore.

CHANGELOG.md
dweinber: Same for L14, or not? We missed that there but would be good to clean up

src/ts/ContentMetadataBuilder.ts
dweinber: Yes, thanks 👍

src/ts/ContentMetadataBuilder.ts
dweinber: Thanks for the comments, makes more sense now 🙂

src/ts/index.ts
dweinber: Ah yes, of course 🤦  The Conviva lib is just a peer dependency, so the customer anyway needs to install it and can access the imports of it. Thanks for clarifying 👍

dweinber (approved):

CHANGELOG.md
karpov-kir: Right, done here https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/106/commits/ecc72d42be7f23f5579bb033c41b7adb0e0dac6c. 🙂

PR #114, opened 2024-08-14
Track initial audio track and subtitles after first play

karpov-kir:
Related to: https://bitmovin.atlassian.net/browse/PA-2533.

This change will allow users to initialize the session externally early, even before `player.load()`.

### DIFF
--- a/src/ts/ConvivaAnalyticsTracker.ts
+++ b/src/ts/ConvivaAnalyticsTracker.ts
@@ -134,6 +134,8 @@ export class ConvivaAnalyticsTracker {
     return this._isAdBreakActive;
   }
 
+  private hasPlayed = false;
+
   /**
    * Do not track play event during ad (e.g. triggered from IMA)
    */
@@ -452,12 +454,6 @@ export class ConvivaAnalyticsTracker {
         Conviva.SystemSettings.LogLevel.ERROR,
       );
     }
-
-    // Send the session init audio language values.
-    this.trackUpdateAudioTrack(this.player.getAudio());
-
-    // Check if at session init has a subtitle enabled.
-    this.trackSubtitleWhenInternalInitialize();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/ts/ConvivaAnalyticsTracker.ts
```diff
@@ -607,6 +604,14 @@ export class ConvivaAnalyticsTracker {
...
       this.internalInitializeSession();
     }
+
+    if (!this.hasPlayed) {
+      this.hasPlayed = true;
+      // Send the session init audio language values.
+      this.trackUpdateAudioTrack(this.player.getAudio());
+      // Check if at session init has a subtitle enabled.
+      this.trackInitialSubtitles();
+    }
```
stonko1994, line 614: I think there is no guarantee that those values are present when the first play event is received or? 🤔 Imagine the use-case where someone calls `player.play()` on bad network before any of this information is available yet?

What if we move this to the `ReadyEvent`? The ready event should always be emitted right **before** playback starts and might be a better place. + it will only be emitted once so we don't need the state handling.

EDIT:
or the first `PlayingEvent` could work as well.

karpov-kir: The current change preserves the initial implementation - https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/blob/v5.0.0/src/ts/ConvivaAnalytics.ts#L639 - here you can see `internalInitializeSession` is invoked inside `Play` event and the initial audio track / subtitles tracking is a part of `internalInitializeSession`.

I guess the `Ready` event does not cover the case if the play was never clicked, so in this case initial audio track / subtitles tracking would be called before the session is initialized (if I understand the sequence of events correctly `Ready` -> `Play` (session initialization) -> `Playing`).

Also, I added:

image

but I don't see any output from the `Ready` event. 🤔

stonko1994 (approved):

src/ts/ConvivaAnalyticsTracker.ts
stonko1994: > I guess the Ready event does not cover the case if the play was never clicked

Good call 🤔 Let's keep as is then 👍

PR #115, opened 2024-08-16
Support late player attaching

karpov-kir:
Related to: 

- https://bitmovin.atlassian.net/browse/TA-2652
- Implementation the same as in https://github.com/bitmovin/bitmovin-player-ios-analytics-conviva/pull/86

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,3 +174,34 @@ The integration will close the active session.
 ```js
 player.load({…});
 ```
+
+#### External VST tracking
+
+If your app needs additional setup steps which should be included in VST tracking, such as DRM token generation, before the `Player` instance can be initialized, 
+the `ConvivaAnalytics` can be initialized without a `Player` instance. Once the `Player` instance is created it can be attached.
+
+1. Create the `ConvivaAnalytics` instance with your `customerKey` and configuration.
+
+```typescript
+const convivaAnalytics = new ConvivaAnalytics(undefined, customerKey);
+```
+
+2. Conviva requires that the `assetName` is set at session initialization. Therefore ensure that you provide using the metadata overrides **before** initializing the tracking session.
+
+```typescript
+convivaAnalytics.updateContentMetadata({
+  assetName: 'Your Asset Name',
+});
+
+// Initialize tracking session
+convivaAnalytics.initializeSession()
+```
+
+3. Once your `Player` instance is ready attach it to the `ConvivaAnalytics` instance.
+
+```typescript
+// ... Additional setup steps
+
+const player = new bitmovin.player.Player(document.getElementById('player'), getPlayerConfig());
+convivaAnalytics.attachPlayer(player);
+``` 
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -48,7 +48,6 @@ export class ConvivaAnalytics {
     this.convivaAnalyticsTracker = new ConvivaAnalyticsTracker(customerKey, config);
     this.debugLoggingEnabled = config.debugLoggingEnabled || false;
     this._player = player;
-    this.handlers = new PlayerEventWrapper(player);
 
     if (player) {
       this.attachPlayer(player);
@@ -106,15 +105,8 @@ export class ConvivaAnalytics {
    * no longer ensure that the session is managed at the correct time.
    */
   public endSession(): void {
-    this.reset();
-    this.convivaAnalyticsTracker.endSession();
-  }
-
-  private reset(): void {
-    this.lastAdBreakEvent = null;
-    this._player = null;
-    this.handlers = null;
     this.convivaSsaiAnalytics.reset();
+    this.convivaAnalyticsTracker.endSession();
   }
 
   /**
@@ -182,14 +174,12 @@ export class ConvivaAnalytics {
   }
 
   public release(): void {
-    this.destroy();
-    this.convivaAnalyticsTracker.release();
-  }
-
-  private destroy(event?: PlayerEventBase): void {
     this.unregisterPlayerEvents();
-    this.reset();
-    this.convivaAnalyticsTracker.release(event);
+    this.lastAdBreakEvent = null;
+    this._player = null;
+    this.handlers = null;
+    this.convivaSsaiAnalytics.reset();
+    this.convivaAnalyticsTracker.release();
   }
 
   private debugLog(message?: any, ...optionalParams: any[]): void {
@@ -318,7 +308,7 @@ export class ConvivaAnalytics {
 
   private onDestroy = (event: any) => {
     this.debugLog('[ ConvivaAnalytics ] [ Player Event ] destroy', event);
-    this.destroy(event);
+    this.release();
   };
 
   private registerPlayerEvents(): void {
--- a/src/ts/ConvivaAnalytics.ts
+++ b/src/ts/ConvivaAnalytics.ts
@@ -19,6 +19,7 @@ import { ConvivaAnalyticsConfiguration, ConvivaAnalyticsTracker, EventAttributes
 import { ConvivaAnalyticsSsai } from './ConvivaAnalyticsSsai';
 import { PlayerEventWrapper } from './helper/PlayerEventWrapper';
 import { AdHelper } from './helper/AdHelper';
+import { Html5Logging } from './Html5Logging';
 
 export class ConvivaAnalytics {
   private handlers?: PlayerEventWrapper;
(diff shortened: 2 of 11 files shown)


src/ts/ConvivaAnalytics.ts
```diff
@@ -1,5 +1,5 @@
 import * as Conviva from '@convivainc/conviva-js-coresdk';
-import type {
+import {
```
stonko1994, line 2: I'm not 100% fluid in TS, but IIRC, there was an issue that when importing events like this, the whole Player was bundled into the final output file (js file) 🤔 Is that no longer the case or is this import not a problem?

src/ts/ConvivaAnalyticsTracker.ts
```diff
@@ -1,16 +1,16 @@
...
   PlaybackEvent,
   PlayerAPI,
-  PlayerEvent,
   PlayerEventBase,
   SourceConfig,
   VideoQualityChangedEvent,
   SubtitleTrack,
   TimeMode,
 } from 'bitmovin-player';
+import { PlayerEvent } from 'bitmovin-player';
```
stonko1994: Why this extra import and not including it in the imports above?

src/ts/ConvivaAnalyticsTracker.ts
```diff
@@ -112,23 +112,34 @@ export interface EventAttributes {
...
+  private get player(): PlayerAPI {
+    if (!this._player) {
+      throw new Error('Player is not initialized, either pass it to the constructor or attach it via `attachPlayer` before using the integration.');
+    }
+    return this._player;
+  }
+
+  private get isPlayerAttached(): boolean {
+    return !!this._player;
+  }
```
stonko1994, line 127: IMO, this approach is a bit dangerous. The fact that the nonoptional player accessor exists (`private get player(): PlayerAPI {`), which will throw an error in case it's not there, makes it easy to forget about the late attaching use-case during development. If someone does a change and accesses `this.player` and forgets about the `isPlayerAttached` check, it will throw an error.

_Also it's easy not to test this use-case if someone is not aware of it._

IMO, having the player optional and seeing that (thanks to the compiler), it can not be forgotten to do either the `null` check or handle the null case explicitly 🤔 

Would like to hear your thoughts on that.

src/ts/ConvivaAnalyticsTracker.ts
```diff
@@ -143,6 +154,52 @@ export class ConvivaAnalyticsTracker {
...
+      reason,
+    };
+  }
+
+  public attachPlayer(player: PlayerAPI): void {
+    const {canAttach} = this.canAttachPlayer(player);
+
+    if (!canAttach) {
+      return;
+    }
```
stonko1994, line 179: This is already verified in the ConvivaAnalytics `attachPlayer` method 🤔 Do we need it here again?

src/ts/ConvivaAnalyticsTracker.ts
```diff
@@ -143,6 +154,52 @@ export class ConvivaAnalyticsTracker {
...
+
+    const playerInfo = {
+      [Conviva.Constants.FRAMEWORK_NAME]: 'Bitmovin Player',
+      [Conviva.Constants.FRAMEWORK_VERSION]: this.player.version,
+    };
+
+    if (this.convivaVideoAnalytics) {
+      this.convivaVideoAnalytics.setPlayerInfo(playerInfo);
+      this.convivaAdAnalytics.setAdPlayerInfo(playerInfo);
+    }
```
stonko1994: I think there is a copy paste error on the first line here (line 188). You check `!this.convivaAdAnalytics` twice. Was is meant to be `convivaVideoAnalytics`? If yes, you can remove the additional if check in line 197

README.md
```diff
@@ -174,3 +174,34 @@ The integration will close the active session.
...
+
+#### External VST tracking
+
+If your app needs additional setup steps which should be included in VST tracking, such as DRM token generation, before the `Player` instance can be initialized, 
+the `ConvivaAnalytics` can be initialized without a `Player` instance. Once the `Player` instance is created it can be attached.
+
+1. Create the `ConvivaAnalytics` instance with your `customerKey` and configuration.
+
+```typescript
+const convivaAnalytics = new ConvivaAnalytics(undefined, customerKey);
```
stonko1994, line 186: Is there a possibility in typescript to omit the `undefined` parameter here? that a customer only need to do `new ConvivaAnalytics(customerKey);`? Or is that not possible as the `player` is the first parameter?

src/ts/ConvivaAnalytics.ts
```diff
@@ -160,15 +204,24 @@ export class ConvivaAnalytics {
     this.convivaAnalyticsTracker.resumeTracking();
   }
 
-  public release(): void {
-    this.destroy();
-    this.convivaAnalyticsTracker.release();
-  }
+  public release(event?: PlayerEventBase): void {
```
stonko1994: I think this change is 'API breaking'? 🤔

src/ts/ConvivaAnalyticsTracker.ts
karpov-kir: The null check is disabled in this repo, so the compiler does not complain. Even if there was the null check enabled, you would need to check that the player exists before accessing it.

`isPlayerAttached` is to pronounce the intent better, instead of just checking `if (this._player)`.

The player is not initialized error should never happen in production. Manual + automated testing should prevent that, hopefully. Without this error, there would be a runtime error anyway in case someone tried to access something on the player instance before it is attached.

karpov-kir: I evaluated if it's easily possible to enable the strict mode, but it will require quite some effort.

src/ts/ConvivaAnalyticsTracker.ts
karpov-kir: This way the relation is more explicit, otherwise you would untangle the implicit link in your head and memorize it, which increases complexity IMO. There is actually a pattern "can execute / execute" I used here.

src/ts/ConvivaAnalytics.ts
karpov-kir: Yeah, I noticed that as well while you were reviewing, here is a fix https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/115/commits/f25bc277b73fade37d6832969805079ae99f089e. 🙂

src/ts/ConvivaAnalyticsTracker.ts
karpov-kir: Because of IDE autoimport, fixed here https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/115/commits/fda346eba962b2411e222bd48b70de119aa0c6ed.

README.md
karpov-kir: There is constructor overloading in TS, but I don't think it's possible in this case:

image

src/ts/ConvivaAnalytics.ts
karpov-kir: You're right, but it's not TS related in this case, but webpack, since webpack bundles the files in this repo. Here is a fix https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/115/commits/8799fd4ffbbb4671b3a8b2418eb0332533649825, https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/115/commits/0416ca5299daa733d85fe402b9a8f4ee0c747b05.

src/ts/ConvivaAnalyticsTracker.ts
karpov-kir: Good catch! Here is a fix https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/115/commits/53107568c71f48e166a01b7226a796dd1dde21fd.

src/ts/ConvivaAnalyticsTracker.ts
```diff
@@ -454,12 +512,22 @@ export class ConvivaAnalyticsTracker {
         Conviva.SystemSettings.LogLevel.ERROR,
       );
     }
+
+    this.setPlayerInfo();
+    // It's required to correctly track VST. There must be BUFFERING or STOPPED metric reported before PLAYING.
+    // In some cases BUFFERING does not fire before PLAYING, so we report STOPPED right after session initialization to cover all edge cases.
+    this.convivaVideoAnalytics.reportPlaybackMetric(Conviva.Constants.Playback.PLAYER_STATE, Conviva.Constants.PlayerState.STOPPED);
```
stonko1994: On mobile we do that **before** we do the `convivaVideoAnalytics.setCallback` setup. I think the idea was that there is no race condition that `convivaVideoAnalytics.setCallback` sneaks in before we report the initial state. Don't know how the timing of the first `setCallback` call is on web.

stonko1994 (approved):

src/ts/ConvivaAnalyticsTracker.ts
karpov-kir: Updated the web implementation to match the mobile https://github.com/bitmovin/bitmovin-player-web-analytics-conviva/pull/115/commits/aaef75b2fcd850375d4e2f257f4e47696993e27e.

PR #116, opened 2024-11-22
Include ad events in stall tracking

wasp898:
Issue: https://bitmovin.atlassian.net/browse/PW-21228

Ad related delays are not contributing to the stalling metrics but should.

### DIFF
--- a/src/ts/ConvivaAnalyticsTracker.ts
+++ b/src/ts/ConvivaAnalyticsTracker.ts
@@ -625,7 +625,7 @@ export class ConvivaAnalyticsTracker {
       PlayerEvent.AdFinished,
     ];
     const stallTrackingClearEvents = [
-      PlayerEvent.StallStarted,
+      PlayerEvent.StallStarted, // StallStarted is reported as BUFFERING immediately. Does not need the delayed timeout approach.
       PlayerEvent.Playing,
       PlayerEvent.Paused,
       PlayerEvent.Seeked,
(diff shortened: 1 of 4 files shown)


src/ts/ConvivaAnalyticsTracker.ts
```diff
@@ -621,15 +621,19 @@ export class ConvivaAnalyticsTracker {
       PlayerEvent.Play,
```
stonko1994, line 621: I can't comment on line 614, but this method is not automatically called for every event. This call still needs to be added accordingly. This call should go into the `ConvivaAnalytics.ts` in the corresponding callback.

However, I tried to do that and couldn't make it work to report buffering to the Ad session. It looks like the player emits the `AdBreakStarted` event only when the VAST and Media Files are resolved (I tried throttling both). Therefore, the buffering is counted towards the main content and not toward the ad 🤔 I am not sure if this is by design. I'm afraid this solution needs more thinking and is not as low-hanging as initially thought.

stonko1994 (changes requested):

src/ts/ConvivaAnalyticsTracker.ts
wasp898: Yeah I suppose this won't actually track the delay in the ad manifest loading, but should at least track successfully the VST of the ad 👀

stonko1994 (approved):
Approving as this seems to be the best we can do for now.

PR #117, opened 2024-11-26
Address underreported VST after ad pre-roll

wasp898:
Issue: https://bitmovin.atlassian.net/browse/PW-21589

When a CSAI pre-roll is configured, the resulting VST of main content ends up being underreported because the player will only dispatch `AdBreakFinished` after main content has fully restored. The startup delay of main content is thus considered as an ad-related delay. 
Since we can't easily change the behavior of `AdBreakFinished` in the player, we will have to work around this via analytics reporting.

Based on how Conviva sessions are defined, our `AdFinished` event is closer to the ad break ended definition than `AdBreakFinished`. With this change, any further delay in playback will contribute to the main content metrics (instead of the ad ones) thus addressing the issue. However, an ad break may include multiple ads, in which case `AdFinished` may be followed by `AdStarted`. For this reason, we can't simply signal end of ad break immediately. Instead we introduce a delay in the signalling so that we can cancel the signalling in case `AdStarted` or `AdBreakFinished` are emitted. If the timer runs out, and we wrongly signal ad break ended, this might lead to some minor misreporting of metrics, but should be minimal overall, while the actual VST issue is still addressed.

A better detection of what to expect from future ads is tricky because APIs like `ads.list()` may not return anything, depending on how IMA is configured. So, designing a more complex workaround is unlikely to be worth it in our opinion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - Ad-related delays not contributing to rebuffering metrics
+- Underreported VST after playing CSAI pre-rolls
 
 ## [5.4.0] - 2024-08-27
 ### Added
--- a/spec/tests/ConvivaAnalyticsTracker.spec.ts
+++ b/spec/tests/ConvivaAnalyticsTracker.spec.ts
@@ -121,6 +121,80 @@ describe(ConvivaAnalyticsTracker, () => {
       expect(stallTrackingStopTimeoutSpy).toHaveBeenCalled();
     })
   })
+
+  /**
+   * Since the web player dispatches AdBreakFinished only after main content has successfully restored,
+   * a workaround is in place to signal the end of the ad break early so that VST can properly be tracked.
+   */
+  describe('eager ad break ended signalling', () => {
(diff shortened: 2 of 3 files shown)


mukulmishra18 (approved):
Looks good to me 😌

mimcz:
Looks good, but I was thinking about another approach, that we could somehow use the `CONTENT_RESUME_REQUESTED` IMA event. instead of the timeout. The problem is we do not expose it directly, we can only hook in using the `onAdsManagerAvailable`. Not sure if we can configure this callback on existing instance. Or we could expose it as new public event in case IMA is used. Not sure if this would be better. Just an idea.

wasp898:
> Looks good, but I was thinking about another approach, that we could somehow use the `CONTENT_RESUME_REQUESTED` IMA event. instead of the timeout. The problem is we do not expose it directly, we can only hook in using the `onAdsManagerAvailable`. Not sure if we can configure this callback on existing instance. Or we could expose it as new public event in case IMA is used. Not sure if this would be better. Just an idea.

As discussed privately, assuming we can do that for lazily attached instances, that would only cover IMA. It makes more sense to either untangle the AdBreakFinished event from main content, or design a brand new event. I think the scope is larger than updating this analytics integration to reflect the current event definitions as best as possible. But we should probably do that nonetheless. 🙂

PR #121, opened 2024-12-23
[PW-22287] Fix delayed buffering reporting for ad-related events

wasp898:
The `stallTrackingTimeout` that is supposed to prevent buffering reporting on in-buffer seeks, was used for unrelated ad events as well. This leads to a small but consistent underreporting.
Furthermore, the introduction of the `RestoringContent` event for CSAI led to expecting this event for SSAI too, which is wrong.

In this PR we split the `stallTrackingTimeout` logic from the regular playback state update reporting, and address the aforementioned signaling issues.


karpov-kir (approved):
🚀

PR #122, opened 2025-10-21
Streamtype override support

RickGroot:
Fixed a bug where the `streamType` wasn't passed correctly to Conviva, passing the source type instead of stream type (by default `dash` was passed as a value instead of `LIVE`/`VOD`/`UNKNOWN`). Also re-added support for an override using the `updateContentMetadata()` function.


dweinber (approved):
I'm not sure why no customer ever brought this up, I'm sure Conviva would catch this in their extensive certification process 🤔 but LGTM

dweinber:
Please fix/correct the tests @RickGroot and we're then good to merge once those are green 👍

PR #125, opened 2026-03-12
feature: Add support for setting custom and required metadata separately

RickGroot:
## Description
This change adds separate update paths for required metadata and custom metadata in the Conviva Bitmovin SDK integration.

Previously, metadata fields with the same name could overwrite each other when required and custom metadata were updated together. By adding the option to update these independently, we avoid unintended overrides and ensure both metadata sets are preserved correctly.

(linting/formatting also made some updates in markdown files)

## Checklist
- [x] `CHANGELOG.md` entry added
- [x] Tests added (or covered by existing ones)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,174 +7,247 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- `updateRequiredContentMetadata` and `updateCustomContentMetadata` methods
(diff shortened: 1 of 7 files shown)


dweinber (approved):
Thanks for your contribution! A good change to have, and nicely done. I like the explicit requiredContentMetadata/customContentMetadata naming and also the Omit/Pick in the types as a nice solution 👌 

I'll leave it to you if you want to address Copilot's minor points above or not.

PR #126, opened 2026-05-18
feature: Capture full error details when reporting playback failures …

priteshtwr:
Description
Summary
Improves field triage. When a playback failure is reported to Conviva, ConvivaAnalyticsTracker.trackError currently forwards only event.code and event.name to reportPlaybackFailed. In production this means operators looking at a Conviva session see something like 1400 SOURCE_ERROR with no indication of which URL failed, what HTTP status came back, or any of the contextual fields the player attaches to event.data — so triaging a stream issue means asking the field team to reproduce and capture player logs.

This PR introduces ErrorHelper.formatPlaybackError and uses it from trackError, so the message reported to Conviva carries the full context the player already has on hand:

code, name, message, type, timestamp
troubleShootLink when present
All known network/HTTP fields from event.data: url, requestUrl, manifestUrl, segmentUrl, httpStatusCode / statusCode / status, statusText, response, responseText, responseUrl, responseHeaders, downloadType, mimeType, method, reason, details, cause, errorMessage, errorCode, subCode, retryCount
Any remaining keys in event.data are JSON-stringified under a data: segment so nothing is silently dropped
Safe stringify handles circular references and BigInt
Example — a 404 on a media segment.

Before:

1400 SOURCE_ERROR

After:

Error code: 1400; Name: SOURCE_ERROR; Message: Segment download failed; Type: error; Timestamp: 1700000000000; url: https://cdn.example.com/segment_1.ts; httpStatusCode: 404; statusText: Not Found; downloadType: media/video;

Nothing about the public API or session lifecycle changes — only the content of the message string passed to reportPlaybackFailed.

### DIFF
--- a/src/ts/helper/ErrorHelper.ts
+++ b/src/ts/helper/ErrorHelper.ts
@@ -1,11 +1,5 @@
 import { ErrorEvent } from 'bitmovin-player';
 
-type ErrorEventWithData = ErrorEvent & {
-  message?: string;
-  troubleShootLink?: string;
-  data?: any;
-};
-
 const KNOWN_DATA_KEYS = [
   'url',
   'requestUrl',
@@ -33,9 +27,10 @@ const KNOWN_DATA_KEYS = [
 
 const URL_LIKE_KEYS = ['url', 'requestUrl', 'manifestUrl', 'segmentUrl', 'responseUrl'];
 const HEADER_LIKE_KEYS = ['responseHeaders'];
-const RESPONSE_BODY_KEYS = ['response', 'responseText'];
 
-// Query-string parameter names that commonly carry credentials in signed-URL schemes.
+// Query-string parameter names that commonly carry credentials in signed-URL
+// schemes — generic OAuth/JWT-style names plus the common CDN signed-URL
+// schemes (CloudFront/S3, Akamai, generic __token__).
 const SENSITIVE_QUERY_PARAMS = [
   'token',
   'access_token',
@@ -53,6 +48,17 @@ const SENSITIVE_QUERY_PARAMS = [
   'sessionid',
   'jwt',
   'bearer',
+  // CloudFront / S3 signed URLs.
+  'x-amz-signature',
+  'x-amz-security-token',
+  'x-amz-credential',
+  'policy',
+  'key-pair-id',
+  'keypairid',
+  // Akamai / generic CDN tokens.
+  'hdnts',
+  'hdnea',
+  '__token__',
 ];
 
 // HTTP header names that commonly carry credentials or session identifiers.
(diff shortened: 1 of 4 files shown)


dweinber:
The commits show up as user @priteshtiwary but the PR was opened and the CLA was signed as @priteshtwr. CLA must be signed by the commit author(s). 
It would be the easiest if @priteshtiwary could sign the CLA, too. Thank you for your understanding.

priteshtiwary:
CLA is signed via correct account now.

priteshtiwary:
Addressed review feedback

1. **`safeStringify` → `undefined`** — coalesce to `String(value)` / `'[Unserializable]'` when `JSON.stringify` returns undefined.
2. **`type` / `timestamp` truthiness** — switched to explicit `!= null` checks so `0` and `''` survive.
3. **Sensitive data** — sanitize on by default: redact known credential query params (`token`, `sig`, `apikey`, …) in URL-like fields, redact sensitive headers (`Authorization`, `Cookie`, `X-Api-Key`, …) in `responseHeaders`, truncate field values >512 chars. Happy to add a config flag as a follow-up if you'd prefer it configurable.
4. **BigInt test** — added.

priteshtiwary:
Hi dweinber,
Are these Co-pilot feedback also must fix?

dweinber:
I'm reviewing the PR and will comment on Copilot's comments after that

src/ts/helper/ErrorHelper.ts
```diff
@@ -0,0 +1,224 @@
...
+      return pair;
+    }
+    const name = pair.slice(0, eq);
+    const val = pair.slice(eq + 1);
+    if (SENSITIVE_QUERY_PARAMS.indexOf(name.toLowerCase()) !== -1) {
+      return `${name}=${REDACTED}`;
+    }
+    return `${name}=${val}`;
+  });
+  return truncate(`${base}?${sanitizedPairs.join('&')}`);
```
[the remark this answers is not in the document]
dweinber: `sanitizeUrl` splits the raw query string on `&`/`=` and compares names against `SENSITIVE_QUERY_PARAMS`. Two gaps:
  - A sensitive value can also live in the path or fragment (...#access_token=secret). Those places are not redacted and would leak.
  - Parameter names like `x-amz-signature` / `Signature=` from S3/CloudFront signed URLs are not in the list. AWS/Akamai/CloudFront signed URLs are the most common real-world signed-URL scheme in video delivery, and their param names (`X-Amz-Signature`, `Policy`, `Key-Pair-Id`, `hdnea`, `__token__`) won't be caught. Worth expanding the list with the common CDN signing params.

src/ts/helper/ErrorHelper.ts
```diff
@@ -0,0 +1,224 @@
+import { ErrorEvent } from 'bitmovin-player';
+
+type ErrorEventWithData = ErrorEvent & {
+  message?: string;
+  troubleShootLink?: string;
+  data?: any;
+};
```
dweinber: Why is this needed? The Bitmovin Player `ErrorEvent` already contains these three fields anyway.

src/ts/helper/ErrorHelper.ts
```diff
@@ -0,0 +1,224 @@
...
+  if (URL_LIKE_KEYS.indexOf(key) !== -1) {
+    return sanitizeUrl(value);
+  }
+  if (HEADER_LIKE_KEYS.indexOf(key) !== -1) {
+    return sanitizeHeaders(value);
+  }
+  if (RESPONSE_BODY_KEYS.indexOf(key) !== -1) {
+    return truncate(safeStringify(value));
+  }
+  return truncate(safeStringify(value));
```
dweinber: As both cases do the same (`return truncate(safeStringify(value));`), we could remove `RESPONSE_BODY_KEYS` and this last `if` here.


```suggestion
  return truncate(safeStringify(value));
```

src/ts/helper/ErrorHelper.ts
```diff
@@ -0,0 +1,224 @@
...
+];
+
+const REDACTED = '[REDACTED]';
+const MAX_FIELD_LENGTH = 512;
+
+function truncate(value: string): string {
+  if (value.length <= MAX_FIELD_LENGTH) {
+    return value;
+  }
+  return `${value.slice(0, MAX_FIELD_LENGTH)}…[truncated ${value.length - MAX_FIELD_LENGTH} chars]`;
```
dweinber: It might be safer to use just regular ASCII dots (`...`) instead of the ellipsis (`…`)

src/ts/helper/ErrorHelper.ts
```diff
@@ -0,0 +1,224 @@
...
+        if (seen.has(val)) {
+          return '[Circular]';
+        }
+        seen.add(val);
+      }
+      if (typeof val === 'bigint') {
+        return val.toString();
+      }
+      return val;
+    });
```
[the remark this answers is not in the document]
dweinber: While valid, this change is a nice-to-have. I don't really expect circular references for this case in the first place.

dweinber (changes requested):
Apologies for the delay. This is a great improvement, thanks for the contribution! I think we're almost there 🙂

priteshtiwary:
@dweinber Addressed the review comments.

dweinber (approved):
Looks good, thank you for your patience and contribution! 🎉

